PR #127, opened 2022-06-27
Fix(CreateTable): Set default standardDefinition

Skraye:
If I want to create a standard table without buffer, cluster or partitioning, I cannot set the standardDefinition field.

This brings a nullPointer since no standardDefinition is provided.


tchiotludo (commented):
fix

PR #156, opened 2022-09-20
Feat(Bigquery.Query): Added destination Table to output and log it

Skraye:
Our users would like to know in which table the query have been loaded without having to go in the flow (also useful when your table name is generated from multiple variables)

edit: I also added the destinationTable in the output so it can be reused in the flow.

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/Query.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/Query.java
@@ -256,7 +256,6 @@ public class Query extends AbstractJob implements RunnableTask<Query.Output>, Qu
                 ),
             this.dryRun
         );
-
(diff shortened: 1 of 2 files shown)


src/main/java/io/kestra/plugin/gcp/bigquery/Query.java
```diff
@@ -426,6 +436,11 @@ public static class Output implements io.kestra.core.models.tasks.Output {
             description = "Only populated if 'store' is set to true."
         )
         private URI uri;
+
+        @Schema(
+            title = "The informations where the data are stored"
+        )
+        private HashMap<String,String> destinationTable;
```
tchiotludo, line 443: Please use a true object for that with properties, it will allow to show documentation on the available one

tchiotludo (changes requested):

PR #214, opened 2023-01-13
Add Firestore tasks

### LINKED ISSUE #87: Firestore

loicmathieu:
Fixes #87


src/main/java/io/kestra/plugin/gcp/firestore/Query.java
```diff
@@ -0,0 +1,230 @@
...
+            title = "Field name for the order by clause."
+    )
+    @PluginProperty(dynamic = false)
+    private String orderBy;
+
+    @Schema(
+            title = "Start offset for pagination of the query results."
+    )
+    @PluginProperty(dynamic = true)
+    private Integer offset;
```
loicmathieu, line 96: @tchiotludo When a dynamic parameter is an Integer we cannot render it with the runContext, will it works this way ? The MongoDB Plugin does the same.

src/main/java/io/kestra/plugin/gcp/firestore/Query.java
```diff
@@ -0,0 +1,230 @@
...
+
+    @Override
+    public Query.Output run(RunContext runContext) throws Exception {
+        try(var firestore = this.connection(runContext)) {
+            var collectionRef = this.collection(runContext, firestore);
+
+            var query = getQuery(runContext, collectionRef, this.queryOperator);
+            if(this.orderBy != null) {
+                query.orderBy(this.orderBy);
+            }
```
loicmathieu, line 112: I'm not a  big fan of those, can we use Optional on the plugin properties or not ? I didn't find any example of how to deal with optional params

src/main/java/io/kestra/plugin/gcp/firestore/List.java
```diff
@@ -0,0 +1,133 @@
...
+                    "collection", this.collection
+            ));
+
+            return output;
+        }
+    }
+
+    private Pair<URI, Long> store(RunContext runContext, Iterable<DocumentReference> documents) throws Exception {
+        File tempFile = runContext.tempFile(".ion").toFile();
+        AtomicLong count = new AtomicLong();
```
loicmathieu, line 84: I reuse AtomicLong counter everywhere as code examples used it (MongoDB plugin for ex) but they should not be needed here as we didn't proceed the list of documents in parallel.

src/main/java/io/kestra/plugin/gcp/firestore/Query.java
tchiotludo: You fell on the dark side of the rendering, we have an issue with int, double, etc ... that can't used `runContext.render` and this can be annoying when using variable. Maybe a ticket on that side to deal it on the future and for now `@PluginProperty(dynamic = false)`

src/main/java/io/kestra/plugin/gcp/firestore/List.java
tchiotludo: It was needed because of lamba, you can increment an `int` on a lambda

src/main/java/io/kestra/plugin/gcp/firestore/Query.java
loicmathieu: OK

src/main/java/io/kestra/plugin/gcp/firestore/List.java
loicmathieu: Oh right, because lambda parameters are final !
I must be tired ...
We can create / find a lighter counter throught, as AtomicLong has some synchronization overhead.

src/main/java/io/kestra/plugin/gcp/firestore/Query.java
loicmathieu: Thinking about it I think we don't have another solution than
- Inject it as String
- Use the runContext to render it
- Parse it
I'll open an issue about this, and we may need to check on all existing plugins that declare a dynamic property and inject it in a primitive wrapper.

PR #219, opened 2023-01-19
feat(pubsub): add Pub/Sub tasks

loicmathieu:
Fixes https://github.com/kestra-io/plugin-gcp/issues/89


src/main/java/io/kestra/plugin/gcp/pubsub/AbstractPubSub.java
```diff
@@ -0,0 +1,64 @@
...
+        VersionProvider versionProvider = runContext.getApplicationContext().getBean(VersionProvider.class);
+
+        TopicName topicName = TopicName.of(runContext.render(projectId), runContext.render(topic));
+        ProjectSubscriptionName subscriptionName = ProjectSubscriptionName.of(runContext.render(projectId), runContext.render(subscription));
+        SubscriptionAdminSettings subscriptionAdminSettings = SubscriptionAdminSettings.newBuilder()
+            .setCredentialsProvider(FixedCredentialsProvider.create(this.credentials(runContext)))
+            .setHeaderProvider(() -> Map.of("user-agent", "Kestra/" + versionProvider.getVersion()))
+            .build();
+
+        // List all existing subscriptions and create the subscription if needed
```
tchiotludo, line 51: We don't need to add a boolean property `createIfNotExists`? 
Maybe some people don't want to auto create subscription?

loicmathieu: Maybe, I'll add a boolean `autoCreateSubscription` with a default value of true

loicmathieu:
@tchiotludo I used a daemon thread as done in the AMQP plugin but I wonder if it's needed for Pub/Sub, I think it'll open it's own polling thread so we may just drop the daemon thread and do everything on the task thread.

tchiotludo:
The daemon thread is only here for blocking code in order to watch duration and rows from the main thread, if pubsub is created a non blocking thread, you can safely remove it

loicmathieu:
@tchiotludo I removed the daemon thread. The PR is ready to go ;)

Issue #265, opened 2023-07-26
Secrets


aballiet:
### Feature description

Currently in order to retrieve secrets hosted on GCP, we need to perform command : 

`gcloud secrets versions access "latest" --secret "MY_SECRET"` using the CLI.

Current flow looks like this :
```YAML
id: retrieve-gcloud-secret
namespace: dev
tasks:
  - id: get-credentials
    type: "io.kestra.plugin.gcp.cli.GCloudCLI"
    serviceAccount: "{{ globals['my-sa-json'] }}"
    commands:
      - gcloud secrets versions access "latest" --secret "MY_SECRETS_ON_GCP" --format json | tr -d '\n ' | xargs -0 -I {} echo '::{"outputs":{"gcloud":{}}}::'
  - id: hello
    type: io.kestra.core.tasks.log.Log
    message: "Your secret value : {{ outputs['get-credentials'].vars['gcloud']['payload']['data'] }}"
```

Aside that the secret value is displayed in the logs, could be interesting to access them more easily

I am willing to do a PR 👍

anna-geller:
closing based on https://github.com/kestra-io/plugin-gcp/pull/266#issuecomment-1725231625

PR #266, opened 2023-07-28
feat: implement gcp secrets read value

aballiet:
First implement of Secret Manager within GCP plugin.
For now, only reading secret value is supported.

FIxes https://github.com/kestra-io/plugin-gcp/issues/265

Need review for testing as it is currentlly WIP


tchiotludo:
We have an internal discussion about this implementation, and this is have a major security breach, all the secret will be exposed in clear in outputs and store inside the execution. 
We have a proper secret management available on the Enterprise Edition and we have made an implementation for google secret that will allow to keep all secrets safe (and never stored).

PR #269, opened 2023-08-16
feat: add Vertex.ai generative AI tasks

loicmathieu:


src/main/java/io/kestra/plugin/gcp/vertexai/ChatCompletion.java
```diff
@@ -0,0 +1,140 @@
...
+        @Example(
+            title = "Chat completion using the Vertex AI PaLM API",
+            code = {
+                """
+                    region: us-central1
+                    projectId: my-project
+                    context: I love jokes that talk about sport
+                    messages:
+                      - author: user
+                        content: Please tell me a joke"""
```
anna-geller, line 46:
```suggestion
                        content: tell me a joke"""
```

you're so polite 😄

anna-geller (approved):
Awesome!

anna-geller:
I only wonder whether we would need to expose `serviceAccount` property? or does it happen automatically? all other GCP tasks and triggers have that

src/main/java/io/kestra/plugin/gcp/vertexai/ChatCompletion.java
loicmathieu: Well, it's a habit I took always to be polite with bots, you never know ... Skynet...

loicmathieu:
@anna-geller 

> I only wonder whether we would need to expose serviceAccount property? or does it happen automatically? all other GCP tasks and triggers have that

It is exposed but not included in the examples as they are not full examples.

Skraye (approved):
Miss some tests
however, LGTM

PR #272, opened 2023-08-21
feat: allow JSON messages for PubSub tasks

### LINKED ISSUE #271: Support JSON for Pub/Sub
Like others queue, we need to provide a `SerdeType` to deserialize JSON easily

loicmathieu:
Fixes #271

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/pubsub/model/Message.java
+++ b/src/main/java/io/kestra/plugin/gcp/pubsub/model/Message.java
@@ -10,6 +10,8 @@ import lombok.Builder;
 import lombok.Getter;
 import lombok.extern.jackson.Jacksonized;
 
+import java.io.IOException;
+import java.util.Base64;
 import java.util.Map;
 
 import static io.kestra.core.utils.Rethrow.throwBiConsumer;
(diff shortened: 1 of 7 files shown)


src/main/java/io/kestra/plugin/gcp/pubsub/model/Message.java
```diff
@@ -35,10 +37,11 @@ public class Message {
     @PluginProperty(dynamic = true)
     private String orderingKey;
 
-    public PubsubMessage to(RunContext runContext) throws IllegalVariableEvaluationException {
+    public PubsubMessage to(RunContext runContext, SerdeType serdeType) throws IllegalVariableEvaluationException, IOException {
         var builder =  PubsubMessage.newBuilder();
         if(data != null) {
-            builder.setData(ByteString.copyFrom(runContext.render(data).getBytes()));
+            var serializedData = serdeType.serialize(data);
```
Skraye, line 43: Data isn't rendered anymore?

loicmathieu: Seems like a bug :(

loicmathieu: Fixed, can you have a look?

Skraye: Sry to be annoying, but can you add the case if people provide a Map<String, Object> and render with [this](https://github.com/kestra-io/kestra/blob/1694df7c26dd66897c76769b479e8382a1864f7b/core/src/main/java/io/kestra/core/runners/RunContext.java#L432-L432), so people can provide object that contains task outputs

loicmathieu: SQS didn't have this functionality neither, this is a different thing.

PR #308, opened 2024-01-05
fix: BigQuery labels was not properly set

aurelienWls:
 BigQuery labels was not properly set

---

By checking that BigQuery Labels are good directly in Google Cloud Logging

---


loicmathieu (approved):
LGTM, thanks

PR #316, opened 2024-01-23
feat(metrics): add BQ total return rows in metrics

aurelienWls:

Add in the Kestra metrics the number of lines return by a BigQuery query.

This is usefull to track overtime the number of lines return by a BigQuery.Query task thanks to Kestra metrics dashboard.

This change involved loading all the data return by the query in the `metrics` method. This could lead to performance issue in case the query return a huge amount of data. This change has not been tested with huge query.

---

Tested locally 

---


loicmathieu:
There is already a `total.rows` metric that return the same metric value (`queryJob.getQueryResults().getTotalRows()` so it seems to be already covered.

This metric will not be displayed if neither store, fetch, or fetchOne is true as in this case the results are not fetched from BQ.

PR #341, opened 2024-03-19
fix(docs): Corrected the multimodal completion task example and minor doc changes

shrutimantri:


src/main/java/io/kestra/plugin/gcp/vertexai/CustomJob.java
```diff
@@ -39,7 +39,7 @@
         @Example(
             code = {
                 "projectId: my-gcp-project",
-                "region: europe-west1",
+                "region: eu-west1",
```
loicmathieu, line 42: The region is `europe-west1` please remove your change

loicmathieu (changes requested):

PR #343, opened 2024-03-19
fix(docs): Correct the GKE ClusterMetadata example

shrutimantri:
- Major change in ClusterMetadata example
- Minor doc corrections


src/main/java/io/kestra/plugin/gcp/gke/ClusterMetadata.java
```diff
@@ -28,35 +28,34 @@
...
     examples = {
         @Example(
-            title = "Fetch a gke cluster metadata",
+            title = "Fetch a GKE cluster's metadata.",
             code = {
-                "name: \"gke-metas\"",
-                "projectId: my-project-id",
-                "zone: eu-west-1c",
+                "clusterProjectId: my-project-id",
+                "clusterZone: eu-west1-c",
```
loicmathieu, line 34: Shouldn't it be `europe-west1-c` instead.

loicmathieu (changes requested):

src/main/java/io/kestra/plugin/gcp/gke/ClusterMetadata.java
shrutimantri: Changed it to "us-central1-a" where it definitely works.
Screenshot 2024-03-19 at 3 45 41 PM

shrutimantri: Corrected. I had actually tried it on "us-central1-a", now tried in "europe-west1-c" as well.

PR #356, opened 2024-03-29
feat(docs): Add metrics to BigQuery tasks

### LINKED ISSUE #353: Document metrics available on BigQuery tasks
### Issue description

Many BQ tasks, such as the Query task, return [multiple metrics](https://kestra.io/blogs/2023-05-30-track-resource-usage-from-bigquery-with-kestra-metric-dashboard) (e.g., total bytes processed). However, those are not documented in the Metrics section as they should be. See the Return task for an example of how metrics are usually documented https://kestra.io/plugins/core/tasks/debugs/io.kestra.core.tasks.debugs.return#metrics

shrutimantri:
Closes #353


loicmathieu (approved):
LGTM

PR #380, opened 2024-05-14
feat: propagate kills to the TaskRunner

fhussonnois:
part-of: kestra-io/kestra#3700


src/main/java/io/kestra/plugin/gcp/runner/GcpBatchTaskRunner.java
```diff
@@ -462,6 +471,34 @@ protected Map<String, Object> runnerAdditionalVars(RunContext runContext, TaskCo
 
         return additionalVars;
     }
+    
+    protected void safelyDeleteJob(final RunContext runContext,
```
loicmathieu, line 475: I would prefer to have `kill` in the method name as there are a lot of lines in the file, and this method is far away from its usage. This would be more informative.
Also, you can make it private but it's nitpicking ;)

loicmathieu (approved):
LGTM

PR #414, opened 2024-07-07
Adjust docs to use MIB units as number instead of GB

japerry911:

- The `2GB` unit for example does not work, one needs to input `2048` as a string (MiB) and it works. Adjusted docs to say this.
- removed the 2 task reference in docs, since Kestra currently does not offer more than one task in GCP Batch job 

- *Note: I don't know if the default is correct in the documentation, as I couldn't find a default set in the code, but it might be a GCP default that is set. Basically this setting of ComputeResources Memory/CPU is required in order to use full power of Batch Runner's compute engine instance. This is because Kestra creates one task for the Compute Engine instance, and you use these settings to set it to use all resources.

---

- Ran it in my Kestra instance in 2 scenarios:
    1. default input parameter values for these settings set in workflow
    2. setting input parameter in a different flow that calls the subflow workflow

```yaml
  - id: cpu
    type: STRING
    defaults: "2000"
    description: Defines the amount of CPU resources used by the trask in milliCPU units.
  
  - id: memory
    type: STRING
    defaults: "4096"
    description: Defines the amount of memory used by the task in MiB units.
```

```yaml
    taskRunner:
      type: io.kestra.plugin.gcp.runner.Batch	
      region: us-central1
      bucket: "{{ secret('GCS_LANDING_BUCKET') }}"
      waitUntilCompletion: 86400
      machineType: "{{ inputs.machine_type }}"
      serviceAccount: "{{ secret('GCP_SERVICE_ACCOUNT_JSON') }}"
      computeResource:
        cpu: "{{ inputs.cpu }}"
        memory: "{{ inputs.memory }}"
```

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/runner/Batch.java
+++ b/src/main/java/io/kestra/plugin/gcp/runner/Batch.java
@@ -543,7 +543,7 @@ public class Batch extends TaskRunner implements GcpInterface, RemoteRunnerInter
 
         @Schema(
             title = "Memory in MiB.",
-            description = "Defines the amount of memory per task in MiB units. If undefined, the default value is `2GB`. " +
+            description = "Defines the amount of memory per task in MiB units. If undefined, the default value is `2048`. " +
                 "If you also define the VM's machine type using the `machineType` in " +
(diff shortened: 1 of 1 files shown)


wrussell1999 (approved):

loicmathieu (approved):
LGTM, AFAIR I get the default from the doc.

PR #434, opened 2024-08-26
feat(docs): add full examples for gcp tasks

shrutimantri:


src/main/java/io/kestra/plugin/gcp/firestore/Set.java
```diff
@@ -28,19 +28,38 @@
...
-            }
+            full = true,
+            code = """
+                id: gcp_firestore_set
+                namespace: company.name
+
+                inputs:
+                  - id: json_string
+                    type: STRING
+                    default: "{\"firstname\": \"John\", \"lastname\": \"Doe\"}"
```
wrussell1999, line 55: Does this need to be escaped inside `"""`? @shrutimantri

shrutimantri: I added one more double quotes around it as there are `:` inside.

wrussell1999 (approved):

Issue #450, opened 2024-10-16
PubSub should not base64 encode the message


loicmathieu:
### Expected Behavior

PubSub should not base64 encode the message

### Actual Behaviour

PubSub Consume and Produce task base64 encode the message.

### Environment Information

- Kestra Version:
- Plugin version: 
- Operating System (OS / Docker / Kubernetes):
- Java Version (If not docker):

PR #451, opened 2024-10-16
fix: don't base64 encode pubsub message

loicmathieu:
Fixes https://github.com/kestra-io/plugin-gcp/issues/450


loicmathieu:
## Breaking change

Pub/Sub messages was previously automatically base64 encoded before being published, and base64 decoded after consumption. This is no longer the case.

PR #459, opened 2024-11-09
feat(gcp): allow ability to impersonate a service account

### LINKED ISSUE #452: Add impersonateServiceAccount option
### Feature description

> Extend the current implementation of credentials for GCP to allow the ability to impersonate service accounts. When using different service accounts to separate permissions, this allo  to create a kestra service account with base permissions and then impersonate other service accounts to run tasks like querying BigQuery. 

```yaml
id: test
namespace: test

tasks:
      - id: query
        impersonateServiceAccount:
        serviceAccount: target@target_project.iam.gserviceaccount.com
            scopes:
                - https://www.googleapis.com/auth/cloud-platform
```

benpops89:
This change allows GCP tasks to be run by impersonating another service account, reasons for this could include
- Enhanced Security with Least Privilege Access
- Delegated and Scoped Access Control
- Temporary Access with Auditing
- Simplifying Resource Access in Multi-Project Environments

Thinking about how to structure this, we could change `serviceAccount` from a `String` to a `HashMap` and have `key` and `impersonate` keys
```yaml
serviceAccount:
    key: "{{ secret('GCP_SERVICE_ACCOUNT') }}"
    impersonate: target-sa@email.com
```

If you do not want to impersonate a service account then do not include the `impersonate` key.

---

```yaml
id: gcp_gcs_list
namespace: company.team

tasks:
  - id: list
    type: io.kestra.plugin.gcp.gcs.List
    serviceAccount:
        key: "{{ secret('GCP_SERVICE_ACCOUNT') }}"
        impersonate: target-sa@email.com
    from: "gs://my_bucket/dir/"
```

I'm not sure of the best way to go about testing this - if guidance can be provided I'm happy to take it on.

---

### Setup Instructions

In order to make this work you will need two service accounts setup and allow the main service account to impersonate the other, documentation on this can be found [here](https://cloud.google.com/iam/docs/service-account-impersonation). Below is an example gcloud command

```shell
gcloud iam service-accounts add-iam-policy-binding TARGET_SERVICE_ACCOUNT_EMAIL \
    --role="roles/iam.serviceAccountTokenCreator" \
    --member="serviceAccount:SOURCE_SERVICE_ACCOUNT_EMAIL" \
    --project=YOUR_PROJECT_ID
```
A good example may be to give one service account no GCS ObjectViewer role but give it to the other one, then the above task would be a good test, as running without impersonate should not work but providing it should then make the task work

I have never touched Java before so this is completely new to me, I think the formatting I'm using may be different to what is being used in the repo as well 🤦.

There will be lots of other places to change references to service account, including any examples, but wanted to get feedback on the concept before tackling that

closes #452

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/CredentialService.java
+++ b/src/main/java/io/kestra/plugin/gcp/CredentialService.java
@@ -45,7 +45,7 @@ public final class CredentialService {
         }
 
         if (gcpInterface.getImpersonatedServiceAccount() != null) {
-            credentials = ImpersonatedCredentials.create(credentials, gcpInterface.getImpersonatedServiceAccount(),
+            credentials = ImpersonatedCredentials.create(credentials, runContext.render(gcpInterface.getImpersonatedServiceAccount()),
                     null,
                     runContext.render(gcpInterface.getScopes()),
                     3600);
--- a/src/main/java/io/kestra/plugin/gcp/CredentialService.java
+++ b/src/main/java/io/kestra/plugin/gcp/CredentialService.java
@@ -11,9 +11,7 @@ import org.slf4j.Logger;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
-import java.util.List;
 import java.util.Map;
-import java.util.HashMap;
 
 public final class CredentialService {
   private CredentialService() {
@@ -23,11 +21,9 @@ public final class CredentialService {
       throws IllegalVariableEvaluationException, IOException {
     GoogleCredentials credentials;
 
-    HashMap<String, String> serviceAccount = gcpInterface.getServiceAccount();
-
-    if (serviceAccount.get("key") != null) {
-      String serviceAccountJson = runContext.render(serviceAccount.get("key"));
-      ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(serviceAccountJson.getBytes());
+    if (gcpInterface.getServiceAccount() != null) {
+      String serviceAccount = runContext.render(gcpInterface.getServiceAccount());
+      ByteArrayInputStream byteArrayInputStream = new ByteArrayInputStream(serviceAccount.getBytes());
       credentials = ServiceAccountCredentials.fromStream(byteArrayInputStream);
       Logger logger = runContext.logger();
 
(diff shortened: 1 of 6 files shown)


loicmathieu (changes requested):
Unfortunatly, we cannot change the serviceAccount property like that, this would be a breaking change.

Maybe we should use two properties:
- serviceAccount: the existing serviceAccount JSON
- impersonateAccount: the impersonation account

benpops89:
@loicmathieu - thanks for the feedback.

I think having a separate property for the impersonated account works well. I'll rework the PR to reflect this.

I've also worked out how to build the develop docker image with local plugins so will be able to test the changes in more detail

benpops89:
@loicmathieu - I have made the requested changes, creating a separate property for impersonated service account.

I have tested by adding the plugin to latest develop docker image and running the following flow, which runs successfully. The service account does not have storage permissions but the impersonated account does

```yaml
id: impersonate
namespace: test

tasks:
  - id: create_bucket
    type: io.kestra.plugin.gcp.gcs.CreateBucket
    projectId: "poppy-demo"
    serviceAccount: "{{ read('demo-sa.json') }}"
    impersonatedServiceAccount: "storage-account@demo.iam.gserviceaccount.com"
    name: "kestra-test-bucket-creation"
```

In order to setup the impersonated account you need to do the following
* Enable the IAM Service Account Credentials API 
* On the service account to be impersonated grant access to the base service account giving it the `Service Account Token Creator` permission. This allows the base service account to create short live tokens for the impersonated account

src/main/java/io/kestra/plugin/gcp/CredentialService.java
```diff
@@ -41,6 +44,13 @@ public static GoogleCredentials credentials(RunContext runContext, GcpInterface
             credentials = credentials.createScoped(runContext.render(gcpInterface.getScopes()));
         }
 
+        if (gcpInterface.getImpersonatedServiceAccount() != null) {
+            credentials = ImpersonatedCredentials.create(credentials, gcpInterface.getImpersonatedServiceAccount(),
```
loicmathieu, line 48: Missing rendering.

```suggestion
            credentials = ImpersonatedCredentials.create(credentials, runContext.render(gcpInterface.getImpersonatedServiceAccount()),
```

loicmathieu (approved):
LGTM, thanks.

Juts a small impl detail, I'll fix it directly myself

PR #483, opened 2025-02-28
Prevent job labels being overwritten for BigQuery

### LINKED ISSUE #482: Add labels to BigQuery QueryJobConfiguration
### Feature description

In BigQuery SDKs you are able to add labels when running queries which could help when reviewing logs. I'm mainly looking to use it to help track costs but assigning labels to then help group costs together, see below how they appear in GCP logs.

Image

The label are just key value pairs and could be added as a new property for BigQuery, the above screenshot was using the Python SDK but I can see from the [documentation](https://cloud.google.com/java/docs/reference/google-cloud-bigquery/latest/com.google.cloud.bigquery.QueryJobConfiguration.Builder#com_google_cloud_bigquery_QueryJobConfiguration_Builder_setLabels_java_util_Map_java_lang_String_java_lang_String__) it's supported in the Java SDK.

I'm going to work on a PR for this as I do not think it's too big a job

benpops89:
For BigQuery job labels ensure the two instances are merged, as currently two separate calls are being made to `builder.setLabels` which causes the user defined labels to be lost, this will close #482

I have made this change to both the `Copy` and `Query` tasks. In addition I have also corrected the type for the `ExtractToGCS` task

---

I ran this against a Google Cloud instance I have setup for Kestra development. It's using the `develop-no-plugins` docker image locally, below is the flow I used to test

```yaml
id: bigquery_labels
namespace: test

tasks:
  - id: labels
    defaultDataset: demo
    labels:
      hello: world
      foo: bar
    sql: |
      SELECT * FROM users
    fetch: true
```
When running the above flow both the user defined and kestra labels are visible (see below)

image

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/Copy.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/Copy.java
@@ -124,15 +124,16 @@ public class Copy extends AbstractJob implements RunnableTask<Copy.Output> {
             builder.setOperationType(runContext.render(this.operationType).as(OperationType.class).orElseThrow().name());
         }
 
-        if (this.labels != null) {
-            builder.setLabels(runContext.render(this.labels).asMap(String.class, String.class ));
-        }
-
         if (this.jobTimeout != null) {
             builder.setJobTimeoutMs(runContext.render(this.jobTimeout).as(Duration.class).orElseThrow().toMillis());
         }
 
-        builder.setLabels(BigQueryService.labels(runContext));
+        Map<String, String> finalLabels = new HashMap<>(BigQueryService.labels(runContext));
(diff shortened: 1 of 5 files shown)


loicmathieu (approved):
LGTM, thanks a lot.

loicmathieu:
@benpops89 there is a compilation error, can you have a look?

benpops89:
@loicmathieu - I'll take a look and see

benpops89:
I've added test cases for `Copy` and `Query` tasks but I'm not sure the best approach for `ExtractToGcs` as it's a bit different to the other two

loicmathieu (approved):
Thanks no need for more tests.

PR #509, opened 2025-06-18
feat: use the new Data.from() facility

loicmathieu:


gradle.properties
```diff
@@ -1,2 +1,2 @@
 version=0.24.0-SNAPSHOT
-kestraVersion=[0.23,)
```
fdelbrayelle, line 2: Why not keep 0.23?

src/main/java/io/kestra/plugin/gcp/pubsub/Publish.java
```diff
@@ -59,12 +50,11 @@
...
 )
-public class Publish extends AbstractPubSub implements RunnableTask<Publish.Output> {
-    @PluginProperty(dynamic = true)
+public class Publish extends AbstractPubSub implements RunnableTask<Publish.Output>, io.kestra.core.models.property.Data.From {
     @NotNull
     @Schema(
-        title = "The source of the published data.",
-        description = "Can be an internal storage URI, a list of Pub/Sub messages, or a single Pub/Sub message.",
+        title = io.kestra.core.models.property.Data.From.TITLE,
+        description = io.kestra.core.models.property.Data.From.DESCRIPTION,
```
fdelbrayelle, line 57: Would it be the same value than before? Not sure to get how it's set here

fdelbrayelle (approved):

PR #511, opened 2025-06-28
fix(bigquery): preserve microsecond precision for DATETIME when fetchType=STORE

### LINKED ISSUE #495: BigQuery fetchType: STORE trims microseconds to miliseconds
### Describe the issue

I have following example of datetime `2025-04-08T13:32:22.87555`

With sqlserver.Query it works well and returns all digits after the dot with fetchType STORE and FETCH
- id: checkAndRecalculateDataOnSQL
    type: "io.kestra.plugin.jdbc.sqlserver.Query"
    url: jdbc:
    username: 
    password: 
    sql: SELECT CAST ('2025-04-08 13:32:22.875550' AS DATETIME2(7)) AS TestDate
    fetchType: STORE

Image

Image

Same behavior I have for bigquery.Query, but only for FETCH, not STORE. With STORE, the result is truncated to 3 digits. 

- id: runRoutine
    sql: |
      SELECT DATETIME ('2025-04-08T13:32:22.875550');
    fetchType: STORE

Image

Image

For us, it is really important, because we don't expect, that some data will be changed during the reading.

### Environment

- Kestra Version: 0.21.7

Malaydewangan09:
When using `fetchType: STORE`, `DATETIME`value was being truncated millisecond precision caused by`java.time.Instant` objects resulted in a lossy conversion.
fixes #495

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/Query.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/Query.java
@@ -629,7 +629,8 @@ public class Query extends AbstractJob implements RunnableTask<Query.Output>, Qu
         }
 
         if (LegacySQLTypeName.DATETIME.equals(field.getType())) {
-            return Instant.parse(value.getStringValue() + "Z");
+            Instant.parse(value.getStringValue() + "Z");
+            return value.getStringValue() + "Z";
         }
 
         if (LegacySQLTypeName.FLOAT.equals(field.getType())) {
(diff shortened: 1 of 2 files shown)


src/main/java/io/kestra/plugin/gcp/bigquery/Query.java
```diff
@@ -629,7 +629,8 @@ private Object convertCell(Field field, FieldValue value, boolean isRepeated) {
         }
 
         if (LegacySQLTypeName.DATETIME.equals(field.getType())) {
-            return Instant.parse(value.getStringValue() + "Z");
+            Instant.parse(value.getStringValue() + "Z");
```
fdelbrayelle, line 632: You can remove this line since you just return a String just behind, else you want to try catch it to do something in case an exception is caught?

Issue #458, opened 2023-10-15
Add a GCP Dataform subplugin


anna-geller:
## Problem

We already support the open-source edition of Dataform:  https://github.com/kestra-io/plugin-dataform 

However, our users requested the ability to trigger Dataform jobs running on GCP Dataform service https://cloud.google.com/dataform?hl=en 

## API

The OSS version was implemented as a Node.js-CLI plugin. However, the GCP-specific plugin will likely only need to talk to GCP Dataform service via the REST API https://cloud.google.com/dataform/reference/rest 

Specifically, the workflow invocation seems like the right endpoint https://cloud.google.com/dataform/reference/rest#rest-resource:-v1beta1.projects.locations.repositories.workflowinvocations 

## Possible syntax

```yaml
id: dataform
namespace: dev
tasks:
    - id: transform
      type: io.kestra.plugin.gcp.dataform.InvokeWorkflow
      wait: true # wait for results by default so that if that job fails, this task fails as well
      # other properties from this request body https://cloud.google.com/dataform/reference/rest/v1beta1/projects.locations.repositories.workflowInvocations#WorkflowInvocation 
```

ideally, we should combine this with the list/get/query endpoints to allow polling for workflow invocation's results (`wait: true`)

j1bulbul:
Hi @anna-geller @Ben8t, I’d like to take on this issue if still relevant/stale?

FYI @MilosPaunovic - not sure if you’re involved here or can confirm but found this on the list you shared :)

fdelbrayelle:
Hi @j1bulbul 👋 

Locally we set a `GOOGLE_SERVICE_ACCOUNT` as shown in the `.github/setup-unit.sh`:

```sh
echo $GOOGLE_SERVICE_ACCOUNT | base64 -d > ~/.gcp-service-account.json
echo "GOOGLE_APPLICATION_CREDENTIALS=$HOME/.gcp-service-account.json" > $GITHUB_ENV
```

And then for the ci this script is executed with a proper service account of us linked to the appropriate project on our own GCP. So you have to use a proper GCP instance of yours to test with a valid service account here.

j1bulbul:
Hi @MilosPaunovic @fdelbrayelle sorry for the late reply. 

I'm currently on holiday but I'll push the PR as soon as I'm back if that's fine.

MilosPaunovic:
Hey @j1bulbul, just checking if we could expect a PR soon?

j1bulbul:
[...]

PR #532, opened 2025-09-17
feat(): add dataform

Malaydewangan09:
closes https://github.com/kestra-io/plugin-gcp/issues/458


src/main/java/io/kestra/plugin/gcp/dataform/AbstractDataForm.java
```diff
@@ -0,0 +1,65 @@
...
+
+        return DataformClient.create(settings);
+
+    }
+protected String buildRepositoryPath(RunContext runContext) throws IllegalVariableEvaluationException {
+    return String.format(
+        "projects/%s/locations/%s/repositories/%s",
+        runContext.render(this.projectId).as(String.class).orElse(null),
+        runContext.render(this.location).as(String.class).orElse(null),
+        runContext.render(this.repositoryId).as(String.class).orElse(null)
```
fdelbrayelle, line 62: ⚠️ This may lead to path like `projects/null/locations/null/repositories/null`

src/main/java/io/kestra/plugin/gcp/dataform/package-info.java
```diff
@@ -0,0 +1,9 @@
+@PluginSubGroup(
```
fdelbrayelle, line 1: Missing icon if we have a sub group

Image)

src/main/java/io/kestra/plugin/gcp/dataform/AbstractDataForm.java
Malaydewangan09: I think better to make these required no?

Screenshot 2025-09-22 at 12 42 18 PM

PR #539, opened 2025-10-03
feat(gcs): trigger based on new file modification/creation

### LINKED ISSUE #537: Google cloud storage Trigger based on new / file modification
The actual [trigger](https://kestra.io/plugins/plugin-gcp/cloud-storage-gcs/io.kestra.plugin.gcp.gcs.trigger) is working only with new file matching the pattern.
Then it's rely on `action`  to move or delete the file. The `NONE` will make the trigger trigger each time (the user add to move, rename the files). 

Some users expect without moving it: 
- To detect only new file matching the pattern
- To detect change of metadata 

We could:
- use the state `runContext.stateStore()` to keep track of metadata. 
- these metadata will list files that capture with metadata (size only ?) 

On the next run, we will remove all the matching previously and not act on them. 

For this, we need a new prop `matching` with : 
- `MATCHING`: actual behavior and default value
- `MATCHING_OR_UPDATED`: the new behavior that will rely on `stateStore`

We could implement then all of these on all files plugins (fs, aws, ...)

Malaydewangan09:
- closes #537 
- closes https://github.com/kestra-io/kestra/issues/11753

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/monitoring/Push.java
+++ b/src/main/java/io/kestra/plugin/gcp/monitoring/Push.java
@@ -58,6 +58,7 @@ public class Push extends AbstractMonitoringTask implements RunnableTask<Push.Ou
     private Property<List<MetricValue>> metrics;
 
     @Schema(title = "The duration window for the metric interval.")
+    @Builder.Default
     private Property<Duration> window = Property.ofValue(Duration.ofMinutes(1));
 
     @Override
--- a/src/main/java/io/kestra/plugin/gcp/monitoring/Query.java
+++ b/src/main/java/io/kestra/plugin/gcp/monitoring/Query.java
@@ -63,12 +63,11 @@ public class Query extends AbstractMonitoringTask implements RunnableTask<Query.
     @Override
     public Output run(RunContext runContext) throws Exception {
         var rFilter = runContext.render(this.filter).as(String.class).orElseThrow();
-        var rWindow = runContext.render(this.window).as(Duration.class).orElse(Duration.ofMinutes(5));
 
         try (MetricServiceClient client = this.connection(runContext)) {
             var rProjectId = runContext.render(this.projectId).as(String.class).orElseThrow();
 
-            Duration duration = runContext.render(this.window)
+            var rWindow = runContext.render(this.window)
                 .as(Duration.class)
                 .orElse(Duration.ofMinutes(5));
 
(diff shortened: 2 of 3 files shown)


src/main/java/io/kestra/plugin/gcp/gcs/Trigger.java
```diff
@@ -93,7 +98,9 @@
         )
     }
 )
-public class Trigger extends AbstractTrigger implements PollingTriggerInterface, TriggerOutput<Downloads.Output>, GcpInterface, ListInterface, ActionInterface{
+public class Trigger extends AbstractTrigger implements PollingTriggerInterface, TriggerOutput<Downloads.Output>, GcpInterface, ListInterface, ActionInterface {
```
fdelbrayelle, line 101: You didn't implement this proposed changed by Anna in the issue to trace the change:

> Add `changeType` (`CREATE` or `UPDATE`) to trigger outputs (non-breaking).

src/test/java/io/kestra/plugin/gcp/gcs/TriggerTest.java
```diff
@@ -189,4 +190,38 @@ void none() throws Exception {
             .build();
         delete.run(TestsUtils.mockRunContext(runContextFactory, task, ImmutableMap.of()));
     }
+
+    @Test
+    void shouldExecuteOnCreate() throws Exception {
```
fdelbrayelle, line 195: Add a test for UPDATE and CREATE_OR_UPDATE too

src/main/java/io/kestra/plugin/gcp/gcs/Trigger.java
```diff
@@ -279,4 +285,26 @@ public static class StateEntry {
...
+        public Blob toBlob() {
+            return blob;
+        }
+    }
+
+
+    @Builder
+    @Getter
+    public static class Output implements io.kestra.core.models.tasks.Output {
+        private final java.util.List<TriggeredBlob> blobs;
```
fdelbrayelle, line 306: Add some doc on this field please 🙏

src/main/java/io/kestra/plugin/gcp/gcs/Trigger.java
```diff
@@ -115,52 +125,193 @@ public class Trigger extends AbstractTrigger implements PollingTriggerInterface,
...
+    record StateContext(boolean flowScoped, String stateName, String stateSubName, String taskRunValue, Optional<Duration> ttl) { }
+
+    @Getter
+    @Builder
+    @AllArgsConstructor
+    @NoArgsConstructor
+    public static class StateEntry {
+        private String uri;
+        private String version;
+        private Instant modifiedAt;
```
fdelbrayelle, line 295: Doc on it too with Anna's specs

Malaydewangan09: But users wont see docs for this, since it is only used for internal mapping🤔

src/main/java/io/kestra/plugin/gcp/gcs/Trigger.java
```diff
@@ -154,6 +154,8 @@ public class Trigger extends AbstractTrigger implements PollingTriggerInterface,
 
     @Override
     public Optional<Execution> evaluate(ConditionContext conditionContext, TriggerContext context) throws Exception {
+
+        System.out.println("running");
```
fdelbrayelle, line 158: To remove

Issue #565, opened 2025-12-01
Refactor `io.kestra.plugin.gcp.function.HttpFunction` to use the new HTTPClient


loicmathieu:
### Describe the issue

Refactor `io.kestra.plugin.gcp.function.HttpFunction` to use the new HTTPClient

Part-of: https://github.com/kestra-io/kestra-ee/issues/4228

### Environment

- Kestra Version: develop

yazanateer:
I would like to take this issue. I will start today

yazanateer:
Thank you! I'm Happy to contribute, excited for more contributions

PR #566, opened 2025-12-03
refactor(http_function): use Kestra HttpClient

yazanateer:
Closes https://github.com/kestra-io/plugin-gcp/issues/565.

This PR refactors the io.kestra.plugin.gcp.function.HttpFunction task to use Kestra’s new internal HTTP client (io.kestra.core.http.client.HttpClient) instead of the deprecated Micronaut HttpClient.
This modernizes the plugin, aligns it with Kestra’s internal standards, and removes the old Netty/Micronaut HTTP dependency.

---

The refactored task was tested using: 

```bash
./gradlew test --tests "io.kestra.plugin.gcp.function.HttpFunctionTest"
```

Results:
Test suite executed successfully
No regressions in HttpFunctionTest
Because GCP tests require Application Default Credentials, only the HttpFunction tests were executed.

---

### Contributor Checklist ✅

- [ ] PR Title and commits follows [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] Add a `closes #ISSUE_ID` or `fixes #ISSUE_ID` in the description if the PR relates to an opened issue.
- [ ] Documentation updated (plugin docs from `@Schema` for properties and outputs, `@Plugin` with examples, `README.md` file with basic knowledge and specifics).
- [ ] Setup instructions included if needed (API keys, accounts, etc.).
- [ ] Prefix all rendered properties by `r` not `rendered` (eg: `rHost`).
- [ ] Use `runContext.logger()` to log enough important infos where it's needed and with the best level (DEBUG, INFO, WARN or ERROR).

⚙️ **Properties**
- [ ] Properties are declared with `Property<T>` carrier type, do **not** use `@PluginProperty`.
- [ ] Mandatory properties must be annotated with `@NotNull` and checked during the rendering.
- [ ] You can model a JSON thanks to a simple `Property<Map<String, Object>>`.

🌐 **HTTP**
- [ ] Must use Kestra’s internal HTTP client from `io.kestra.core.http.client`

📦 **JSON**
- [ ] If you are serializing response from an external API, you may have to add a `@JsonIgnoreProperties(ignoreUnknown = true)` at the mapped class level. So that we will avoid to crash the plugin if the provider add a new field suddenly.
- [ ] Must use Jackson mappers provided by core (`io.kestra.core.serializers`)

✨ **New plugins / subplugins**
- [ ] Make sure your new plugin is configured like mentioned [here](https://kestra.io/docs/plugin-developer-guide/gradle#mandatory-configuration).
- [ ] Add a `package-info.java` under each sub package respecting [this format](https://github.com/kestra-io/plugin-odoo/blob/main/src/main/java/io/kestra/plugin/odoo/package-info.java) and choosing the right category.
- [ ] Every time you use `runContext.metric(...)` you have to add a `@Metric` ([see this doc](https://kestra.io/docs/plugin-developer-guide/document#document-the-plugin-metrics))
- [ ] Docs don't support to have both tasks/triggers in the root package (e.g. `io.kestra.plugin.kubernetes`) and in a sub package (e.g. `io.kestra.plugin.kubernetes.kubectl`), whether it's: all tasks/triggers in the root package OR only tasks/triggers in sub packages.
- [ ] Icons added in `src/main/resources/icons` in SVG format and not in thumbnail (keep it big):
  - `plugin-icon.svg`
  - One icon per package, e.g. `io.kestra.plugin.aws.svg`
  - For subpackages, e.g. `io.kestra.plugin.aws.s3`, add `io.kestra.plugin.aws.s3.svg`
    See example [here](https://github.com/kestra-io/plugin-elasticsearch/blob/master/src/main/java/io/kestra/plugin/elasticsearch/Search.java#L76).
- [ ] Use `"{{ secret('YOUR_SECRET') }}"` in the examples for sensible infos such as an API KEY.
- [ ] If you are fetching data (one, many or too many), you must add a `Property<FetchType> fetchType` to be able to use `FETCH_ONE`, `FETCH` and even `STORE` to store big amount of data in the internal storage.
- [ ] Align the `"""` to close examples blocks with the flow id.
- [ ] Update the existing `index.yaml` for the main plugin, and for each new subpackage add a metadata file named exactly after the subpackage (e.g. `s3.yaml` for `io.kestra.plugin.aws.s3`) under `src/main/resources/metadata/`, following the same schema.

🧪 **Tests**
- [ ] Unit Tests added or updated to cover the change (using the `RunContext` to actually run tasks).
- [ ] Add sanity checks if possible with a YAML flow inside `src/test/resources/flows`.
- [ ] Avoid disabling tests for CI. Instead, configure a local environment whenever it's possible with `.github/setup-unit.sh` (to be set executable with `chmod +x setup-unit.sh`) (which can be executed locally and in the CI) all along with a new `docker-compose-ci.yml` file (do **not** edit the existing `docker-compose.yml`). If needed, create an executable (`chmod +x cleanup-unit.sh`) `cleanup-unit.sh` to remove the potential costly resources (tables, datasets, etc).
- [ ] Provide screenshots from your QA / tests locally in the PR description. The goal here is to use the JAR of the plugin and directly test it locally in Kestra UI to ensure it integrates well.

📤 **Outputs**
- [ ] Do not send back as outputs the same infos you already have in your properties.
- [ ] If you do not have any output use `VoidOutput`.
- [ ] Do not output twice the same infos (eg: a status code, an error code saying the same thing...).

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/function/HttpFunction.java
+++ b/src/main/java/io/kestra/plugin/gcp/function/HttpFunction.java
@@ -3,25 +3,24 @@ package io.kestra.plugin.gcp.function;
 import com.google.auth.oauth2.IdTokenCredentials;
 import com.google.auth.oauth2.ServiceAccountCredentials;
 import io.kestra.core.exceptions.IllegalVariableEvaluationException;
-import io.kestra.core.http.client.configurations.TimeoutConfiguration;
+import io.kestra.core.http.HttpRequest;
+import io.kestra.core.http.HttpResponse;
+import io.kestra.core.http.client.HttpClient;
+import io.kestra.core.http.client.HttpClientResponseException;
 import io.kestra.core.http.client.configurations.HttpConfiguration;
+import io.kestra.core.http.client.configurations.TimeoutConfiguration;
 import io.kestra.core.models.annotations.Example;
 import io.kestra.core.models.annotations.Plugin;
 import io.kestra.core.models.property.Property;
 import io.kestra.core.models.tasks.RunnableTask;
 import io.kestra.core.runners.RunContext;
-import io.kestra.plugin.gcp.AbstractTask;
-
-import io.kestra.core.http.HttpRequest;
-import io.kestra.core.http.HttpResponse;
-import io.kestra.core.http.client.HttpClient;
-import io.kestra.core.http.client.HttpClientResponseException;
 import io.kestra.core.serializers.JacksonMapper;
-
+import io.kestra.plugin.gcp.AbstractTask;
 import io.swagger.v3.oas.annotations.media.Schema;
 import jakarta.validation.constraints.NotNull;
 import lombok.*;
 import lombok.experimental.SuperBuilder;
+import org.slf4j.Logger;
 
 import java.net.URI;
 import java.time.Duration;
@@ -53,6 +52,7 @@ import java.util.Map;
     )
 })
 public class HttpFunction extends AbstractTask implements RunnableTask<HttpFunction.Output> {
+
     @Schema(title = "HTTP method")
     @NotNull
     protected Property<String> httpMethod;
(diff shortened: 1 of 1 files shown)


src/main/java/io/kestra/plugin/gcp/function/HttpFunction.java
```diff
@@ -93,39 +79,51 @@ public Output run(RunContext runContext) throws Exception {
...
-            HttpResponse result =  maxDuration != null ? mono.block(runContext.render(maxDuration).as(Duration.class).orElseThrow()) : mono.block();
-            String body = result != null &&  result.getBody().isPresent() ? (String) result.getBody().get() : "";
-            try {
-                ObjectMapper mapper = new ObjectMapper();
+        //resolving dynamic properties
+        String rMethod = runContext.render(this.httpMethod).as(String.class).orElseThrow();
+        String rUrl = runContext.render(this.url).as(String.class).orElseThrow();
+        Map<String, Object> rBodyMap = runContext.render(this.httpBody).asMap(String.class, Object.class);
+
+        try(HttpClient client = new HttpClient(runContext, null)) {
```
fdelbrayelle:
```suggestion
        try (var client = new HttpClient(
            runContext,
            HttpConfiguration.builder()
                .timeout(
                    TimeoutConfiguration.builder()
                        .readIdleTimeout(Duration.ofSeconds(60))
                        .build()
                ).build()
        )) {
```

yazanateer: resolved but instead of .readIdleTimeout(Duration.ofSeconds(60)) it should be                             .readIdleTimeout(Property.ofValue(Duration.ofSeconds(60)))
becuase in the TimeoutConfiguration file the readIdleTimeout is declared as Property

yazanateer:
I have pushed the updated implementation, the PR is ready for review
Thanks

fdelbrayelle (approved):

PR #624, opened 2026-06-11
fix(gcs): default Trigger/Downloads action to NONE

jymaire:
## Root cause

`io.kestra.plugin.gcp.gcs.Trigger` and `io.kestra.plugin.gcp.gcs.Downloads` both call:

```java
runContext.render(action).as(Action.class).orElseThrow()
```

but their `action` property had **no default value**. When a flow did not configure `action`, the render produced an empty `Optional` and `.orElseThrow()` threw `NoSuchElementException` with the message `No value present`.

For the **Trigger**, this code path is only reached *after* the empty-result guard (`toFire.isEmpty()`), i.e. once new/updated files have actually been detected. So the trigger failed at the exact moment it was supposed to fire, with a worker log like:

```
WARN  Trigger evaluation failed in the worker with error: No value present
ERROR No value present
```

The `@Schema` description already documented `NONE` as the default ("NONE (default), MOVE, or DELETE"), so the behavior was a code/doc mismatch rather than an intended requirement.

## Fix

- Default `action` to `Action.NONE` (via `@Builder.Default`, matching the existing convention for `interval`/`listingType`/`scopes`) in both `Trigger` and `Downloads`.
- Update the `Downloads` schema description to mention `NONE (default)`.
- Add a regression test `TriggerTest#shouldFireWhenActionNotConfigured` that builds a `Trigger` with no `action`, detects an uploaded file, and asserts evaluation does not throw and fires with the blob present.

## Impact

This is reported by a production user (upgrade to 0.22) who had to work around it by switching to the deprecated `gcs.LegacyTrigger` with an explicit `action: NONE`. This fix lets them use `gcs.Trigger` without setting `action`, enabling migration off the legacy implementation.

## Testing

- `compileJava` / `compileTestJava` pass.
- The new and existing trigger tests require live GCP credentials (`@EnabledIfEnvironmentVariable("GOOGLE_APPLICATION_CREDENTIALS")`) and a configured bucket, so they run in CI rather than locally.

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/gcs/Downloads.java
+++ b/src/main/java/io/kestra/plugin/gcp/gcs/Downloads.java
@@ -104,9 +104,10 @@ public class Downloads extends AbstractGcs implements RunnableTask<Downloads.Out
 
     @Schema(
         title = "Post-download action",
-        description = "DELETE or MOVE (copy to moveDirectory then delete source)"
+        description = "NONE (default), DELETE, or MOVE (copy to moveDirectory then delete source)"
     )
-    private Property<ActionInterface.Action> action;
+    @Builder.Default
+    private final Property<ActionInterface.Action> action = Property.ofValue(ActionInterface.Action.NONE);
 
     @Schema(
         title = "Move destination",
--- a/src/main/java/io/kestra/plugin/gcp/gcs/Trigger.java
+++ b/src/main/java/io/kestra/plugin/gcp/gcs/Trigger.java
@@ -120,7 +120,8 @@ public class Trigger extends AbstractTrigger implements PollingTriggerInterface,
         title = "Post-detection action",
         description = "NONE (default), MOVE (requires moveDirectory), or DELETE on source objects after download"
     )
-    private Property<ActionInterface.Action> action;
+    @Builder.Default
+    private final Property<ActionInterface.Action> action = Property.ofValue(ActionInterface.Action.NONE);
 
     @Schema(
         title = "Move destination",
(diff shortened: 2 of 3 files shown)


jymaire:
## QA against real GCS

Ran the trigger tests against a live GCS bucket (`kestra-unit-test`) using application-default credentials, not mocks. Tests are gated by `@EnabledIfEnvironmentVariable("GOOGLE_APPLICATION_CREDENTIALS")` and `serviceAccount` is unset, so `CredentialService` falls back to `GoogleCredentials.getApplicationDefault()`.

### With the fix
| Test | Result |
|---|---|
| `shouldFireWhenActionNotConfigured` (new regression) | ✅ PASS |
| `none` | ✅ PASS |
| `shouldExecuteOnCreate` | ✅ PASS |

`tests=3, skipped=0, failures=0, errors=0` — no skips, so credentials/bucket access were genuinely live.

### Without the fix (sanity check)
Reverted `action` back to no-default and re-ran the regression test → ❌ fails with `action: must not be null`, confirming the test actually guards the bug.

### Note on the two failure modes
The buggy path surfaces differently depending on the layer:
- **Test harness (`mockTrigger`)** → `jakarta.validation.ConstraintViolationException: action: must not be null` (the `@NotNull` on `ActionInterface.getAction()` is bean-validated at build time).
- **Customer's running worker** → `NoSuchElementException: "No value present"` from `.orElseThrow()` (an already-deployed flow is evaluated without re-running that bean validation, so the empty `Optional` reaches `performAction`).

Same root cause — no default for `action` — and defaulting to `NONE` fixes both. This also explains why the customer's pre-existing flow loaded fine but only failed at evaluation time once files were detected.

Malaydewangan09 (approved):

Issue #622, opened 2026-06-10
Add BigTable, Spanner, Dataflow, and RCS Business Messaging sub-plugins to plugin-gcp


fdelbrayelle:
## Summary

Consolidate four GCP service integrations from the plugin backlog into the `plugin-gcp` repository: Bigtable (NoSQL wide-column store), Spanner (distributed relational database), Dataflow (managed Apache Beam), and RCS Business Messaging (rich mobile messaging via Google RBM). These four sub-plugins cover high-value GCP services already in production use at many teams running on GCP, making `plugin-gcp` a more comprehensive platform integration for orchestrated workflows.

## Plugins in Scope

- **`io.kestra.plugin.gcp.bigtable`** — Read, write, and delete rows in Google Cloud Bigtable; polling trigger for new rows
- **`io.kestra.plugin.gcp.spanner`** — Execute SQL queries and DML against Cloud Spanner databases; trigger via Change Streams
- **`io.kestra.plugin.gcp.dataflow`** — Launch Classic and Flex Template Dataflow jobs, monitor job state, cancel jobs; trigger on job completion
- **`io.kestra.plugin.gcp.rcs`** — Send RCS messages, Rich Cards, carousels, and file attachments via Google RCS Business Messaging (RBM)

## Acceptance Criteria

- [ ] All four sub-issues listed above are implemented, reviewed, and merged
- [ ] Each sub-package has its own `package-info.java`, `metadata/<name>.yaml`, and icon SVG
- [ ] `AGENTS.md` updated with all new task/trigger class names
- [ ] Integration documented in the Kestra docs

fdelbrayelle:
Do you want to take the whole EPIC @suraj-0628 ?

suraj-0628:
Yeahh Sure I'll complete this epic 🚀

fdelbrayelle:
Thanks!

PR #636, opened 2026-06-23
feat(gcp): add Google Cloud Spanner sub-plugin 

### LINKED ISSUE #86: Spanner
https://cloud.google.com/spanner

suraj-0628:
## Summary
Implements the `spanner` sub-plugin under `io.kestra.plugin.gcp.spanner` as part of the GCP plugin backlog consolidation (#622).
Enables Kestra workflows to interact with Google Cloud Spanner — the globally consistent, horizontally scalable relational database service — using the same credential management, retry logic, and output tracking as existing GCP sub-plugins.
## Tasks & Triggers Added
| Class | Description |
|---|---|
| `AbstractSpanner` | Abstract base with shared GCP auth (projectId, instanceId, databaseId, credentials via `CredentialService`) |
| `Query` | Execute SQL queries with parameter binding; supports `FETCH`, `FETCH_ONE`, and `STORE` |
| `Execute` | Execute single transactional DML statements or database schema DDL changes |
| `BatchDml` | Execute multiple DML statements atomically in a single transaction |
| `CreateDatabase` | Create a Spanner database with optional initial DDL statements |
| `DeleteDatabase` | Drop/delete a database permanently |
| `Trigger` | Change stream polling trigger detecting data modifications (inserts, updates, deletes) via two-step partition token discovery |
## Plugin Structure
- Package: `io.kestra.plugin.gcp.spanner`
- `package-info.java` with `@PluginSubGroup(category = {CLOUD, DATA})`
- `metadata/spanner.yaml` and `io.kestra.plugin.gcp.spanner.svg` added
- `doc/io.kestra.plugin.gcp.spanner.md` added
- `AGENTS.md` updated with new task class names
## Coding Standards Met
- [x] All properties use `Property<T>` — no legacy `@PluginProperty(dynamic = true)`
- [x] Secret/credential properties annotated with `@PluginProperty(secret = true)`
- [x] Every property and output annotated with `@Schema` (without trailing periods in titles)
- [x] Five Lombok annotations on all task classes (`@SuperBuilder`, `@ToString`, `@EqualsAndHashCode`, `@Getter`, `@NoArgsConstructor`)
- [x] Logging via `runContext.logger()` only
- [x] JSON serialization via Jackson mappers from `io.kestra.core.serializers`
- [x] GCP auth via `CredentialService` (same pattern as `bigquery`, `gcs`)
- [x] `@Plugin(examples = ...)` entries with `full = true` and `{{ secret(...) }}` for sensitive values
## Gradle Dependency Added
```groovy
api 'com.google.cloud:google-cloud-spanner:6.67.0'
```
## Testing

- [x] `./gradlew build` passes
- [x] `./gradlew test` passes
- Manual QA validated against real Kestra instance using shadow jar + Docker

## Related

Closes part of EPIC #622
Closes https://github.com/kestra-io/plugin-gcp/issues/86


jymaire (changes requested):
Hello @suraj-0628 and thanks for the PR!

Here are some review comments: 

🚫 Blocking issues

  1. SQL injection — changeStreamName is concatenated directly into Change Stream SQL in
  Trigger#evaluate / discoverPartitionTokens with no validation. Spanner TVF names can't be bind
  params, so an @Pattern identifier constraint is required.
  2. Unbounded trigger output — every change record is accumulated into an in-JVM List placed
  directly in the (persisted, non-streamed) trigger output → OOM risk on busy tables. Should use
  internal storage + a URI, scalar counts only.
  3. Per-partition client churn — a new Spanner gRPC client (thread + channel pools) is created
  per partition token per poll cycle. Create one client outside the loop.
  4. Debug logs at INFO — leftover logger().info() calls in Trigger#executeQuery log per-row
  internals every poll. Remove.
  5. Fragile DDL detection in Execute — startsWith("create"/"alter"/"drop") misses
  RENAME/GRANT/REVOKE and breaks on leading block comments. Prefer an explicit isDdl property or a
  proper prefix set.

  ⚠️  Request-changes (conventions/tests)

  - Test gaps: no FETCH/STORE/DDL coverage; @BeforeAll catch blocks swallow exceptions
  - Hardcoded task IDs + missing @Execution(SAME_THREAD); Thread.sleep(2000) → use Await
  - collect(Collectors.toList()) ×7 → .toList(); missing var usage throughout

Feel free to ask more details if some points are too high level.

suraj-0628:
Hi @jymaire 👋 , thanks for the review, looking into it !

jymaire (approved):

src/main/resources/icons/io.kestra.plugin.gcp.spanner.svg
jymaire, line 1: please same icon as https://www.g2.com/fr/products/google-cloud-spanner/reviews

src/main/java/io/kestra/plugin/gcp/spanner/AbstractSpanner.java
```diff
@@ -0,0 +1,238 @@
...
+                    binder.to(Value.float64Array(list.stream().mapToDouble(o -> ((Number) o).doubleValue()).toArray()));
+                } else {
+                    binder.to(Value.stringArray(list.stream().map(Object::toString).toList()));
+                }
+            }
+        } else {
+            try {
+                String json = JacksonMapper.ofJson().writeValueAsString(value);
+                binder.to(Value.json(json));
+            } catch (Exception e) {
```
jymaire: try to reduce scope of Exception (a more specific one)

src/main/java/io/kestra/plugin/gcp/spanner/AbstractSpanner.java
```diff
@@ -0,0 +1,238 @@
...
+            binder.to(com.google.cloud.Timestamp.ofTimeSecondsAndNanos(((java.time.Instant) value).getEpochSecond(), ((java.time.Instant) value).getNano()));
+        } else if (value instanceof java.time.LocalDate) {
+            java.time.LocalDate ld = (java.time.LocalDate) value;
+            binder.to(com.google.cloud.Date.fromYearMonthDay(ld.getYear(), ld.getMonthValue(), ld.getDayOfMonth()));
+        } else if (value instanceof byte[]) {
+            binder.to(Value.bytes(com.google.cloud.ByteArray.copyFrom((byte[]) value)));
+        } else if (value instanceof List) {
+            var list = (List<?>) value;
+            if (list.isEmpty()) {
+                binder.to(Value.stringArray(null));
```
jymaire: **Bug: empty list binds a wrong-typed NULL array.**

`Value.stringArray(null)` is not an empty array — it is the SQL value `NULL` typed as `ARRAY<STRING>`. Two problems:

1. **NULL ≠ empty.** With `WHERE id IN UNNEST(@ids)`, an empty array yields 0 rows (correct), but a NULL array makes the predicate evaluate to `NULL` (3-valued logic) and breaks anything like `ARRAY_LENGTH(@ids)`.
2. **Wrong element type.** It forces `STRING` regardless of the target column. For an `INT64` column this hard-fails:
   ```
   FAILED_PRECONDITION: No matching signature for operator IN UNNEST
     for argument types: INT64, ARRAY<STRING>
   ```

You genuinely can't infer the element type from an empty list, so silently defaulting to `ARRAY<STRING>` NULL is the worst option — it fails far from the cause. Reject it loudly instead (or expose a typed-array parameter API):

```java
if (list.isEmpty()) {
    throw new IllegalArgumentException(
        "Cannot bind an empty list for parameter '" + name + "': "
        + "Spanner arrays are typed and the element type cannot be inferred from an empty list. "
        + "Omit the parameter or pass a typed/non-empty value.");
}
```

src/main/java/io/kestra/plugin/gcp/spanner/AbstractSpanner.java
```diff
@@ -0,0 +1,238 @@
...
+                if (first instanceof String) {
+                    binder.to(Value.stringArray(list.stream().map(o -> (String) o).toList()));
+                } else if (first instanceof Boolean) {
+                    binder.to(Value.boolArray(list.stream().map(o -> (Boolean) o).toList()));
+                } else if (first instanceof Long || first instanceof Integer) {
+                    binder.to(Value.int64Array(list.stream().mapToLong(o -> ((Number) o).longValue()).toArray()));
+                } else if (first instanceof Double || first instanceof Float) {
+                    binder.to(Value.float64Array(list.stream().mapToDouble(o -> ((Number) o).doubleValue()).toArray()));
+                } else {
+                    binder.to(Value.stringArray(list.stream().map(Object::toString).toList()));
```
jymaire: **Bug: array element type inferred from `list.get(0)` only.** Three failure modes:

1. **NPE on a null element.** `[1L, null, 3L]` picks the `int64Array` branch, then `((Number) o).longValue()` over `null` throws an NPE with no useful message. Note the scalar path (line 104) handles `null`, but a `null` *inside a list* does not — inconsistent.
2. **`Short`/`Byte` lists take the wrong type.** The scalar branch (line 113) accepts `Short`/`Byte` as `INT64`, but here only `Long`/`Integer` are checked, so a `List<Short>` falls through to the `else` and becomes `ARRAY<STRING>`. A `Short` scalar → `INT64`, a `Short` list → `ARRAY<STRING>`. The two paths disagree on what "integer" means.
3. **Heterogeneous lists key off the wrong element.** `[1L, 2.5]` picks `int64Array` and silently truncates `2.5` → `2`; `[2.5, 1L]` picks `float64Array`. Same logical data, different result depending on order.

A robust version validates the list is homogeneous, rejects nulls, and treats `Short`/`Byte` like the scalar path. Sketch:

```java
} else if (value instanceof List<?> list) {
    bindList(binder, name, list);
}
...
private void bindList(ValueBinder<Statement.Builder> binder, String name, List<?> list) {
    if (list.isEmpty()) { /* see comment on empty-list handling */ }
    if (list.stream().anyMatch(Objects::isNull)) {
        throw new IllegalArgumentException("Null elements are not supported in array parameter '" + name + "'");
    }
    Object first = list.get(0);
    switch (first) {
        case String s   -> binder.to(Value.stringArray(cast(list, String.class, name)));
        case Boolean b  -> binder.to(Value.boolArray(cast(list, Boolean.class, name)));
        case Long l     -> binder.to(Value.int64Array(toLongs(list, name)));
        case Integer i  -> binder.to(Value.int64Array(toLongs(list, name)));
        case Short sh   -> binder.to(Value.int64Array(toLongs(list, name)));   // consistent with scalar
        case Byte by    -> binder.to(Value.int64Array(toLongs(list, name)));
        case Double d   -> binder.to(Value.float64Array(toDoubles(list, name)));
        case Float f    -> binder.to(Value.float64Array(toDoubles(list, name)));
        default         -> binder.to(Value.stringArray(list.stream().map(Object::toString).toList()));
    }
}
```
where `cast`/`toLongs`/`toDoubles` assert every element matches the first element's type and throw a clear error on a mixed list, instead of truncating silently.

jymaire (commented):
Two correctness issues in `bindParameter`'s array handling. Both come from the same root cause: a Spanner `ARRAY<T>` parameter's element type is part of its type signature and can't be reliably derived from the runtime contents of a `List`. Details inline.

jymaire (changes requested):
sorry, found other bugs after first approval

suraj-0628:
@jymaire sure will resolve it  ; )

src/main/java/io/kestra/plugin/gcp/spanner/AbstractSpanner.java
```diff
@@ -0,0 +1,238 @@
...
+        } else if (value instanceof com.google.cloud.Timestamp) {
+            binder.to((com.google.cloud.Timestamp) value);
+        } else if (value instanceof com.google.cloud.Date) {
+            binder.to((com.google.cloud.Date) value);
+        } else if (value instanceof java.time.Instant) {
+            binder.to(com.google.cloud.Timestamp.ofTimeSecondsAndNanos(((java.time.Instant) value).getEpochSecond(), ((java.time.Instant) value).getNano()));
+        } else if (value instanceof java.time.LocalDate) {
+            java.time.LocalDate ld = (java.time.LocalDate) value;
+            binder.to(com.google.cloud.Date.fromYearMonthDay(ld.getYear(), ld.getMonthValue(), ld.getDayOfMonth()));
+        } else if (value instanceof byte[]) {
```
jymaire: **Design: convert this `instanceof` cascade to a Java 21 pattern `switch`.** The project targets Java 21 (`build.gradle`), and `getValue`/`getArrayValue` in this same class already use `switch` — so this cascade is inconsistent with the author's own style two methods down. It also folds the `value == null` early-return (line 104) into a `case null`:

```java
switch (value) {
    case null              -> binder.to(Value.string(null));
    case String s          -> binder.to(Value.string(s));
    case Boolean b         -> binder.to(Value.bool(b));
    case Long l            -> binder.to(Value.int64(l));
    case Integer i         -> binder.to(Value.int64(i.longValue()));
    case Short sh          -> binder.to(Value.int64(sh.longValue()));
    case Byte by           -> binder.to(Value.int64(by.longValue()));
    case Double d          -> binder.to(Value.float64(d));
    case Float f           -> binder.to(Value.float64(f.doubleValue()));
    case BigDecimal bd     -> binder.to(Value.numeric(bd));
    case Timestamp ts      -> binder.to(ts);
    case Date dt           -> binder.to(dt);
    case Instant inst      -> binder.to(Timestamp.ofTimeSecondsAndNanos(inst.getEpochSecond(), inst.getNano()));
    case LocalDate ld      -> binder.to(Date.fromYearMonthDay(ld.getYear(), ld.getMonthValue(), ld.getDayOfMonth()));
    case byte[] bytes      -> binder.to(Value.bytes(ByteArray.copyFrom(bytes)));
    case List<?> list      -> bindList(binder, name, list);   // see array-bug comments
    default                -> bindJson(binder, value);
}
```
`case null` must come first (a `switch` NPEs on a null selector otherwise).

src/main/java/io/kestra/plugin/gcp/spanner/AbstractSpanner.java
```diff
@@ -0,0 +1,238 @@
+package io.kestra.plugin.gcp.spanner;
+
+import java.io.IOException;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.LinkedHashMap;
+import java.util.ArrayList;
+import java.util.stream.Collectors;
```
jymaire: **Unused imports.** `java.util.ArrayList` (line 8) and `java.util.stream.Collectors` (line 9) are never used here (`rowToMap` uses `LinkedHashMap`, and the streams use `.toList()`). A build with `-Werror` would fail on these.

src/main/java/io/kestra/plugin/gcp/spanner/BatchDml.java
```diff
@@ -0,0 +1,94 @@
+package io.kestra.plugin.gcp.spanner;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
```
jymaire: **Unused import** — `java.util.stream.Collectors` is not used (the mapping uses `.toList()`).

jymaire (commented):
Follow-up review: design, consistency, and test-coverage findings (non-blocking, in addition to the two array bugs already flagged).

src/main/java/io/kestra/plugin/gcp/spanner/BatchDml.java
```diff
@@ -0,0 +1,94 @@
...
+                """
+        )
+    }
+)
+public class BatchDml extends AbstractSpanner implements RunnableTask<BatchDml.Output> {
+
+    @NotNull
+    @Schema(title = "The list of SQL DML statements to execute")
+    @PluginProperty(group = "main")
+    private Property<List<String>> statements;
```
jymaire, line 58: **Design: no parameterization.** `Query` and `Execute` support bound `parameters` via `bindParameter`, but `BatchDml` takes only raw strings and uses `Statement.of(...)` directly. That means the task most likely to be fed templated/interpolated values has no safe binding path, pushing users toward string concatenation. Consider supporting per-statement parameters, or at minimum document this limitation in the schema description.

src/main/java/io/kestra/plugin/gcp/spanner/Trigger.java
```diff
@@ -0,0 +1,337 @@
...
+                }
+            }
+        }
+        return new AbstractMap.SimpleEntry<>(localRowCount, localChangeCount);
+    }
+
+    @SuperBuilder
+    @NoArgsConstructor
+    public static class SpannerClientFactory extends AbstractSpanner {
+    }
```
jymaire: **Design: this empty subclass is a code smell, and it duplicates the connection fields.** `SpannerClientFactory extends AbstractSpanner {}` exists only to reach the `protected` helpers (`spannerClient`, `bindParameter`, `rowToMap`, `databaseId`), and `Trigger` re-declares all 7 connection fields (lines 68-99) that already live on `AbstractSpanner`, copying them into the factory via the builder.

This repo's convention (see `pubsub/Trigger.java:117-129`) is for a trigger to build the underlying task and call `.run()`. The clean fix here: extract the reusable logic into a **static `SpannerService`** utility that both `AbstractSpanner` and `Trigger` call, and factor the shared connection fields into a `SpannerConnectionInterface` (mirroring `PubSubConnectionInterface`). That removes both the field duplication and the empty subclass.

src/main/java/io/kestra/plugin/gcp/spanner/Trigger.java
```diff
@@ -0,0 +1,337 @@
...
+                        if (earliestTimestamp != null) {
+                            finalStartTimestamp = earliestTimestamp;
+                            partitionTokens = discoverPartitionTokens(runContext, dbClient, clientFactory, rChangeStreamName, finalStartTimestamp, endTimestamp);
+                        } else {
+                            throw e;
+                        }
+                    } else {
+                        throw e;
+                    }
+                }
```
jymaire: **Duplicated retry logic.** This `catch (SpannerException) -> OUT_OF_RANGE -> parseEarliestTimestamp -> retry` block is repeated almost verbatim at lines 196-213. Extract a helper (e.g. `runWithEarliestTimestampRetry(...)`) so the recovery logic lives in one place. `evaluate()` is ~110 lines; this is the easiest way to shorten it.

src/main/java/io/kestra/plugin/gcp/spanner/Trigger.java
```diff
@@ -0,0 +1,337 @@
...
+                var earliestInstant = Instant.parse(tsStr);
+                var earliestTimestamp = com.google.cloud.Timestamp.ofTimeSecondsAndNanos(earliestInstant.getEpochSecond(), earliestInstant.getNano());
+                logger.info("Provided startTimestamp was before stream creation. Retrying query with earliest read timestamp: {}", earliestTimestamp);
+                return earliestTimestamp;
+            } catch (Exception ex) {
+                logger.error("Failed to parse earliest read timestamp: {}", tsStr, ex);
+            }
+        }
+        return null;
+    }
```
jymaire, line 266: **Brittle: parsing a timestamp out of a human-readable error message.** `indexOf("earliest read timestamp:")` couples behavior to the exact wording of the Spanner client's error string. If Google changes that message, this silently returns `null` and the original exception is rethrown — recovery just stops working with no signal. At minimum add a prominent comment documenting the coupling; ideally derive the earliest timestamp from a typed source rather than string-scraping.

src/main/java/io/kestra/plugin/gcp/spanner/Trigger.java
```diff
@@ -0,0 +1,337 @@
...
+    @PluginProperty(group = "advanced")
+    private Property<String> emulatorHost;
+
+    @NotNull
+    @Schema(title = "The Change Stream name to poll")
+    @PluginProperty(group = "source")
+    private Property<String> changeStreamName;
+
+    @Schema(
+        title = "Lookback window window applied to poll stream records.",
```
jymaire: **Typo** in the schema title: "Lookback window window applied..." — duplicated "window".

src/main/java/io/kestra/plugin/gcp/spanner/Trigger.java
```diff
@@ -0,0 +1,337 @@
...
+                }
+
+                logger.info("Discovered Spanner change stream partition tokens: {}", partitionTokens);
+
+                for (var token : partitionTokens) {
+                    var querySql = "SELECT * FROM READ_" + rChangeStreamName + "(" +
+                        "start_timestamp => @startTimestamp, " +
+                        "end_timestamp => @endTimestamp, " +
+                        "partition_token => @partitionToken, " +
+                        "heartbeat_milliseconds => 10000" +
```
jymaire: **Magic number.** `heartbeat_milliseconds => 10000` is hardcoded here and again at line 241. Pull it into a named constant (or make it configurable) so the two call sites can't drift.

src/main/java/io/kestra/plugin/gcp/spanner/Trigger.java
```diff
@@ -0,0 +1,337 @@
...
+                        } else {
+                            throw e;
+                        }
+                    }
+                }
+            }
+        }
+
+        if (rowCount == 0 || changeCount == 0) {
+            tempFile.delete();
```
jymaire: **Ignored return value.** `tempFile.delete()` returns a `boolean` that's discarded; most linters flag this. Either check it (and log on failure) or use `Files.deleteIfExists(tempFile.toPath())`. Also note `finalStartTimestamp` (line 151) is named `final*` but is reassigned at line 167 — misleading name.

src/main/java/io/kestra/plugin/gcp/spanner/Query.java
```diff
@@ -0,0 +1,155 @@
...
+    private Property<Map<String, Object>> parameters;
+
+    @Builder.Default
+    @Schema(title = "Result handling mode")
+    @PluginProperty(group = "processing")
+    private Property<FetchType> fetchType = Property.ofValue(FetchType.NONE);
+
+    @Override
+    public Output run(RunContext runContext) throws Exception {
+        String rSql = runContext.render(this.sql).as(String.class).orElseThrow();
```
jymaire: **Consistency nit:** `Query` and `AbstractSpanner` use explicit types throughout, while `Execute`, `BatchDml`, and `Trigger` use `var`. Both styles exist in the repo, but please pick one *within this new package* — right now it reads like two different authors.

src/test/java/io/kestra/plugin/gcp/spanner/SpannerTest.java
```diff
@@ -0,0 +1,341 @@
...
+        RunContext runContext = runContextFactory.of();
+
+        Execute executeTask = Execute.builder()
+            .id("execute")
+            .type(Execute.class.getName())
+            .projectId(Property.ofValue(PROJECT_ID))
+            .instanceId(Property.ofValue(INSTANCE_ID))
+            .databaseId(Property.ofValue(DATABASE_ID))
+            .sql(Property.ofValue("INSERT INTO users (id, name, age) VALUES (@id, @name, @age)"))
+            .parameters(Property.ofValue(Map.of("id", 1L, "name", "John Doe", "age", 30L)))
```
jymaire, line 147: **Biggest test gap: `bindParameter` is the most complex code in the PR and is barely covered.** Only `String`/`Long` are exercised here. There is no coverage for `Boolean`, `Double`, `BigDecimal`, `byte[]`, `Timestamp`/`Instant`, `Date`/`LocalDate`, the JSON fallback, or — most importantly given the two array bugs flagged — **arrays and empty lists**. Please add a parameterized round-trip test (bind → query back → assert value *and* type) per supported type, plus the array edge cases (empty list, `List<Short>`, list containing a null, heterogeneous list).

Also missing: a negative test asserting that an invalid `changeStreamName` is rejected (the regex guard at `Trigger.java:129` is a security fix with no test). And `createAndDeleteDatabaseTask` uses `System.currentTimeMillis()` for the DB name (line 271) — prefer a fixed name with guaranteed teardown for reproducibility.

jymaire (approved):

jymaire:
> @jymaire sure will resolve it ; )

Thanks a lot @suraj-0628 GG

jymaire:
@suraj-0628 deployed in https://github.com/kestra-io/plugin-gcp/releases/tag/v2.7.0

PR #638, opened 2026-06-25
feat(dataflow): add Google Cloud Dataflow sub-plugin

### LINKED ISSUE #88: Dataflow

suraj-0628:
## Summary

Implements the `dataflow` sub-plugin under `io.kestra.plugin.gcp.dataflow` enabling Kestra workflows to launch, monitor, and cancel Google Cloud Dataflow Classic and Flex Template jobs.
Enables orchestration of Dataflow processing pipelines directly within Kestra flows, utilizing native credential management, execution state monitoring, and metrics tracking.

## Tasks & Triggers Added

| Class | Description |
|---|---|
| `AbstractDataflow` | Abstract base with shared GCP connection parameters (`projectId`, `serviceAccount`, `impersonatedServiceAccount`, `scopes`, `location` via `CredentialService` and `DataflowConnectionInterface`) |
| `LaunchTemplate` | Launch a Classic Template Dataflow job from a Cloud Storage path with runtime parameter mapping |
| `LaunchFlexTemplate` | Launch a Flex Template Dataflow job from a spec JSON file with container spec variables mapping |
| `GetJob` | Retrieve current status, timestamps, job type, and execution metrics of a Dataflow job by ID |
| `CancelJob` | Cancel or drain a running Dataflow job, allowing streaming pipelines to finalize buffer data |
| `WaitForJob` | Poll job execution state sequentially using `Await.until` until it enters a terminal state |
| `Trigger` | Polling trigger firing workflow executions upon detecting matching job prefix state transitions |

## Plugin Structure

- Package: `io.kestra.plugin.gcp.dataflow`
- `package-info.java` with `@PluginSubGroup(categories = {PluginSubGroup.PluginCategory.CLOUD})`
- `metadata/dataflow.yaml` and `io.kestra.plugin.gcp.dataflow.svg` added
- `doc/io.kestra.plugin.gcp.dataflow.md` added
- `AGENTS.md` updated with new task class names

## Coding Standards Met

- [x] All properties use `Property<T>` — no legacy `@PluginProperty(dynamic = true)`
- [x] Secret/credential properties annotated with `@PluginProperty(secret = true)`
- [x] Every property and output annotated with `@Schema` (without trailing periods in titles)
- [x] Five Lombok annotations on all task classes (`@SuperBuilder`, `@ToString`, `@EqualsAndHashCode`, `@Getter`, `@NoArgsConstructor`)
- [x] Logging via `runContext.logger()` only
- [x] JSON serialization via Jackson mappers from `io.kestra.core.serializers`
- [x] GCP auth via `CredentialService` (same pattern as `bigquery`, `gcs`)
- [x] `@Plugin(examples = ...)` entries with `full = true` and `{{ secret(...) }}` for sensitive values
- [x] Clean files with absolutely zero comments in code
- [x] Java 21 `var` variable type standards met

## Gradle Dependency Added

```groovy
implementation 'com.google.apis:google-api-services-dataflow:v1b3-rev20240430-2.0.0'
```
## Testing

- [x] `./gradlew build` passes
- [x] `./gradlew test` passes
- Manual QA validated against real Kestra instance using shadow jar + Docker

## Related

Closes part of EPIC #622
Closes https://github.com/kestra-io/plugin-gcp/issues/88


src/main/java/io/kestra/plugin/gcp/dataflow/AbstractDataflow.java
```diff
@@ -0,0 +1,53 @@
...
+@ToString
+@EqualsAndHashCode(callSuper = true)
+@Getter
+@NoArgsConstructor
+public abstract class AbstractDataflow extends Task implements GcpInterface, DataflowConnectionInterface {
+
+    @NotNull
+    @Schema(title = "The GCP project ID")
+    @PluginProperty(group = "connection")
+    protected Property<String> projectId;
```
Malaydewangan09: You can extend AbstractTask (_src/main/java/io/kestra/plugin/gcp/AbstractTask_) also here?

src/main/java/io/kestra/plugin/gcp/dataflow/DataflowService.java
```diff
@@ -0,0 +1,24 @@
...
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
+import com.google.api.client.json.gson.GsonFactory;
+import com.google.api.services.dataflow.Dataflow;
+import com.google.auth.http.HttpCredentialsAdapter;
+
+import io.kestra.core.runners.RunContext;
+import io.kestra.plugin.gcp.CredentialService;
+
+public class DataflowService {
+    public static Dataflow dataflowClient(RunContext runContext, DataflowConnectionInterface connection) throws Exception {
```
Malaydewangan09: why not move this to Abstractdataflow?

src/main/java/io/kestra/plugin/gcp/dataflow/Trigger.java
```diff
@@ -0,0 +1,178 @@
...
+
+    @NotNull
+    @Schema(title = "The Dataflow job name prefix to match")
+    @PluginProperty(group = "source")
+    private Property<String> jobNamePrefix;
+
+    @Builder.Default
+    @Schema(title = "The target state to trigger on (e.g. JOB_STATE_DONE, JOB_STATE_FAILED)")
+    @PluginProperty(group = "processing")
+    private Property<String> targetState = Property.ofValue("JOB_STATE_DONE");
```
Malaydewangan09: Is this list exhaustive, if yes, then lets extract this to an ENUM

src/main/java/io/kestra/plugin/gcp/dataflow/AbstractDataflow.java
```diff
@@ -0,0 +1,53 @@
...
+    @PluginProperty(secret = true, group = "connection")
+    protected Property<String> serviceAccount;
+
+    @Schema(title = "The GCP service account to impersonate")
+    @PluginProperty(secret = true, group = "advanced")
+    protected Property<String> impersonatedServiceAccount;
+
+    @Schema(title = "The GCP scopes to be used")
+    @PluginProperty(group = "advanced")
+    protected Property<List<String>> scopes;
```
Malaydewangan09:
```suggestion
    protected Property<List<String>> scopes = Property.ofValue(List.of("https://www.googleapis.com/auth/cloud-platform"));
```

src/main/java/io/kestra/plugin/gcp/dataflow/DataflowService.java
Malaydewangan09: if this is not used anywhere else

fdelbrayelle:
## Summary

This PR adds a `dataflow` sub-plugin with six classes: `LaunchTemplate`, `LaunchFlexTemplate`, `GetJob`, `CancelJob`, `WaitForJob`, and a polling `Trigger`. The structure, annotations, and test approach are generally correct. Several issues need to be addressed before merge, ranging from a correctness bug in the trigger to a missing `index.yaml` entry and a `@PluginProperty(group)` inconsistency inherited from `GcpInterface`.

---

## Issues

### Correctness

- **[BLOCK] `Trigger.java:756-758` — Dataflow Jobs list is not paginated**
  The `listResponse` call returns at most one page of results (default page size ~50 on the Dataflow API). If there are more jobs in the region the trigger silently misses them. The loop must follow `listResponse.getNextPageToken()` until `null`. Without pagination, the trigger will randomly miss completions in busy projects and appear to work correctly only in low-volume environments.

- **[BLOCK] `LaunchTemplate.java:604`, `LaunchFlexTemplate.java:469` — NPE on `response.getJob()` when API returns null**
  Both `launch` endpoints can return a response where `getJob()` is `null` (e.g. if the job submission was accepted but the API did not include a job object in the response). `var job = response.getJob()` is dereferenced unconditionally. Add an explicit null-check and throw an actionable exception with the request details.

- **[REQUEST CHANGES] `WaitForJob.java:925-930` — redundant second null-check is dead code**
  `Await.until` already throws `TimeoutException` when the condition never returns a non-null value before the deadline. The `if (finalJob == null)` block after the catch can never be reached — `Await.until` either returns the value or throws, it does not return `null` on timeout. Remove the dead branch; it adds false confidence that a second code path is guarded.

- **[REQUEST CHANGES] `WaitForJob.java:934-939` — throw `Exception` is too broad**
  Both failure branches throw bare `Exception`. Use `IllegalStateException` (or a dedicated Kestra exception) with a message that tells the user what to do: `"Dataflow job 'X' failed — check the Dataflow console for error details"`. Throwing `Exception` forces callers to catch everything.

### Kestra Guidelines

- **[REQUEST CHANGES] `AbstractDataflow.java` / `Trigger.java` — `serviceAccount` group mismatches `GcpInterface`**
  `GcpInterface` declares `serviceAccount` with `@PluginProperty(secret = true, group = "execution")`. `AbstractDataflow` overrides the annotation with `group = "connection"`, and `Trigger` does the same. This creates two different group assignments for the identical property depending on which path renders the UI, and diverges from the contract defined in the shared interface. Either remove the per-class `@PluginProperty` annotation (let `GcpInterface` annotation win) or align to `"execution"` consistently everywhere.

- **[REQUEST CHANGES] `Trigger.java` — `@EqualsAndHashCode` without `callSuper = true`**
  The `Trigger` class annotates with `@EqualsAndHashCode` (no `callSuper`), while `AbstractDataflow` and all task classes correctly use `@EqualsAndHashCode(callSuper = true)`. For triggers, the parent `AbstractTrigger` holds the trigger `id` and `type`, which must be included in equality for deduplication to work correctly in the Kestra trigger evaluation loop. Change to `@EqualsAndHashCode(callSuper = true)` to match the convention used in `spanner/Trigger.java`.

- **[REQUEST CHANGES] `DataflowService.java` — dependency should be `api`, not `implementation`**
  All other GCP client libraries in `build.gradle` are declared as `api`. Declaring `google-api-services-dataflow` as `implementation` means consumers of this plugin (via shadow-jar delegation) cannot access the Dataflow model classes directly. Use `api` for consistency and to avoid class-not-found surprises in downstream integrations.

- **[REQUEST CHANGES] `Trigger.java:725` — `targetState` should be a typed enum, not `Property<String>`**
  The valid values (`JOB_STATE_DONE`, `JOB_STATE_FAILED`, etc.) are a closed set. A `Property<String>` gives no validation and no UI autocompletion. Define a `JobState` enum (or reuse the string set from `WaitForJob.isTerminalState`) and use `Property<JobState>`. This also eliminates the silent bug where a user typo like `"JOB_STATE_DONE "` would never trigger.

- **[SUGGESTION] `metadata/dataflow.yaml` — missing `videos: []` field**
  All other `metadata/*.yaml` files (pubsub, bigquery, spanner, etc.) include `videos: []`. The `dataflow.yaml` omits it. While not a hard error, it deviates from the schema convention every other subpackage follows.

- **[SUGGESTION] `index.yaml` — not updated**
  The main `src/main/resources/metadata/index.yaml` was not updated to reference the new `dataflow` subpackage. Other subpackages are referenced through their own yaml files but the convention of updating `index.yaml` when adding a new subpackage should be verified against project norms.

### Security

- **[REQUEST CHANGES] `GetJob.java:310`, `WaitForJob.java:953` — broad `Exception` caught in metrics fallback**
  Catching `Exception` swallows `RuntimeException` subclasses (e.g. `IllegalArgumentException`, `NullPointerException`) that signal programming errors, not transient API failures. The intended catch target is `IOException` (the checked exception from `.execute()`). Change to `catch (IOException e)` in both metrics-fetch catch blocks so real bugs are not silently swallowed.

- **[SUGGESTION] `DataflowService.java` — no HTTP timeout configured on the transport**
  `GoogleNetHttpTransport.newTrustedTransport()` uses default timeouts. For a polling-heavy plugin where jobs can take hours, a hung HTTP connection will block a Kestra worker thread indefinitely. Set an explicit connect and read timeout via `NetHttpTransport.Builder` with reasonable defaults (e.g. 30s connect / 60s read), or document that the default is acceptable.

### Performance

- **[REQUEST CHANGES] `GetJob.java:298`, `WaitForJob.java:941` — `HashMap` used for metrics output**
  `metrics` is built as a mutable `HashMap` and then stored in an immutable `Output` record. Use `Map.copyOf(metricsMap)` before passing to the builder to prevent unintended mutation and align with the immutability preference stated in plugin standards.

- **[SUGGESTION] `WaitForJob.java` — new `Dataflow` client created once but re-used across poll iterations inside `Await.until`**
  This is fine as written (the client is created before the `Await.until` block and reused). No change needed, noting it is correct.

---

## Suggestions

- `CancelJob.java` — the `requestedState` output echoes an input-derived value. Per output design guidelines, skip redundant outputs unless downstream tasks genuinely need them templated. A `currentState` from the API response would be more useful than the pre-computed `requestedState` string the task itself chose.

- `Trigger.java` — the trigger fires only on the **first** matching job in the list that falls within the lookback window. If two jobs with the same name prefix both completed in the window, only one execution is generated and the other is permanently dropped. Document this behaviour explicitly in the `@Schema(description)`, or change the loop to return the most-recently-completed match.

- `DataflowConnectionInterface.java` — this interface just re-declares the same methods already on `GcpInterface` plus `getLocation()`. Since `AbstractDataflow` already implements both `GcpInterface` and `DataflowConnectionInterface`, the only additive method is `getLocation()`. Consider collapsing `DataflowConnectionInterface` to declare only `getLocation()` (extending `GcpInterface`) to reduce boilerplate.

- Test `DataflowTest.java` — `MockitoAnnotations.openMocks(this).close()` in `@BeforeEach` immediately closes the `AutoCloseable` returned by `openMocks`. This closes the mock-resource container before the test runs. Call `.close()` in a `@AfterEach` instead to avoid subtle mock lifecycle issues.

fdelbrayelle:
You have conflicts to resolve @suraj-0628

suraj-0628:
sorry for that @fdelbrayelle

fdelbrayelle:
Did you take all our review feedback into account @suraj-0628 ?

suraj-0628:
@fdelbrayelle  Yes I believe 🙏🏻

fdelbrayelle:
## Re-review of the cumulative diff (head `1f8cb41`)

I re-read the full diff, the two prior review threads (the detailed review body from `@fdelbrayelle` and the inline comments from `@Malaydewangan09`), and re-checked each previously flagged item against the current code, plus ran a fresh pass for anything not caught before.

### Verdict: **REQUEST CHANGES**

Every item raised in the previous two review rounds has been fixed. Two new, concrete issues surfaced in this pass that should be fixed before merge (see "New findings"). Nothing here blocks the overall design; the fixes are localized.

---

### Previously flagged issues — all confirmed fixed

| # | Issue | Status |
|---|---|---|
| 1 | `Trigger` job list not paginated | **Fixed** — `Trigger.java:136-162` now loops on `getNextPageToken()` until `null`. |
| 2 | NPE on `response.getJob()` in `LaunchTemplate`/`LaunchFlexTemplate` | **Fixed** — both now null-check and throw `IllegalStateException` with the response payload in the message. |
| 3 | `WaitForJob` dead null-check after `Await.until` | **Fixed** — removed; `finalJob` is used directly after the try/catch. |
| 4 | `WaitForJob` throwing bare `Exception` | **Fixed** — `WaitForJob.java:124-128` now throws `IllegalStateException` with actionable messages ("check the Google Cloud Dataflow console for error details"). |
| 5 | `serviceAccount` group mismatch (`connection` vs `execution` from `GcpInterface`) | **Fixed** — `AbstractDataflow` no longer redeclares the field (extends `AbstractTask`, inherits `GcpInterface`'s `group = "execution"`); `Trigger.java:76-78` explicitly aligned to `group = "execution"`. |
| 6 | `Trigger` `@EqualsAndHashCode` without `callSuper` | **Fixed** — `Trigger.java:35` now `@EqualsAndHashCode(callSuper = true)`. |
| 7 | Dataflow client lib declared as `implementation` | **Fixed** — `build.gradle:85` now `api 'com.google.apis:google-api-services-dataflow:...'`. |
| 8 | `targetState` as free `Property<String>` | **Fixed** — `JobState` enum added (`JobState.java`), `Trigger.java:99-101` now `Property<JobState>`. |
| 9 | `metadata/dataflow.yaml` missing `videos: []` | **Fixed**. |
| 10 | Broad `catch (Exception e)` around metrics fetch | **Fixed** — both `GetJob.java:308` and `WaitForJob.java:143` now `catch (IOException e)`. |
| 11 | No HTTP timeout on the Dataflow transport | **Fixed** — `DataflowService.java` sets `setConnectTimeout(30000)` / `setReadTimeout(60000)` on the request initializer. |
| 12 | Mutable `HashMap` returned as output | **Fixed** — both `GetJob` and `WaitForJob` now build with `Map.copyOf(metricsMap)`. |
| 13 | `CancelJob` echoing input-derived `requestedState` as output | **Fixed** — output now returns `currentState` from the actual API response. |
| 14 | Trigger firing only on first match, ignoring later matches in the same window | **Fixed** — `Trigger.java:133-162` tracks `mostRecentJob`/`mostRecentTime` across the full paginated result set; behavior documented in the class `@Schema` description. |
| 15 | `DataflowConnectionInterface` boilerplate duplicating `GcpInterface` | **Fixed** — collapsed to just `getLocation()`, extends `GcpInterface`. |
| 16 | `MockitoAnnotations.openMocks(this).close()` closed immediately in `@BeforeEach` | **Fixed** — `DataflowTest.java` now stores the `AutoCloseable` and closes it in a proper `@AfterEach`. |

The "update `index.yaml`" suggestion from the previous round doesn't apply here — checked the existing convention (`spanner.yaml`, `pubsub.yaml`, etc.) and the root `metadata/index.yaml` never references sub-package yaml files in this repo, so there's nothing to update. Not a real gap.

---

### New findings (not previously raised)

**🔴 [Correctness] `Trigger.java:118-176` — no persisted state across poll cycles, risking duplicate or missed executions**

The trigger recomputes a sliding `[now - lookback, now)` window from scratch on every poll (`Trigger.java:126-129`) and has no memory of which job/state-transition it already fired on — unlike the sibling `io.kestra.plugin.gcp.gcs.Trigger`, which uses `StatefulTriggerService` to persist a version/timestamp per object specifically to avoid re-triggering.

Concrete failure modes:
- If `lookback` is set larger than `interval` (a reasonable choice to guard against an occasionally slow/skipped poll — the whole reason `lookback` exists as a separate knob), the same completed job will match in *every* poll whose window still covers its `stateTime`, firing one duplicate execution per poll until it ages out.
- If `lookback` is left unset (defaults to `interval`, `Trigger.java:126`) and a single poll is delayed or skipped (worker busy, GC pause, scheduler backlog), the event falls outside the next window and is silently missed forever.

This is exactly the "remove or advance processed items to avoid infinite re-triggering" requirement for `PollingTriggerInterface`. Fix: persist the last-fired `stateTime` (or job ID) per trigger (e.g. via `StatefulTriggerService`, same pattern as `gcs.Trigger`) and only fire for a job whose `stateTime` is strictly after the last persisted value, instead of re-deriving the window from `Instant.now()` each time.

**🟠 [Reliability] `WaitForJob.java` (and by extension `LaunchTemplate`/`LaunchFlexTemplate`) — no `kill()`/`stop()` override, orphaning the Dataflow job on execution kill**

`WaitForJob.run()` (lines 88-152) submits a poll loop that can run for up to `maxDuration` (default 1h, unbounded) via `Await.until`. If the Kestra execution is killed (SLA timeout, manual kill) while this is polling, the worker thread stops but nothing cancels or drains the actual Dataflow job — it keeps running (and billing, potentially across many worker VMs) with no link back to Kestra anymore.

This is the exact defect class called out in this repo's own recent history: `AbstractBigquery.kill()`/`stop()` (see `io.kestra.plugin.gcp.bigquery.AbstractBigquery`, and `io.kestra.plugin.gcp.vertexai.CustomJob`) were added specifically to fix "BigQuery job tasks didn't cancel the underlying BQ job on kill, so it kept consuming slots and billing after the execution was already KILLED" (#657, merged just before this PR). `RunnableTask` ships default no-op `kill()`/`stop()` from `WorkerJobLifecycle`; none of the five Dataflow tasks override them.

Fix: on `WaitForJob` at minimum, track the job id + `Dataflow` client behind an `AtomicReference<Runnable> killable` / `AtomicBoolean isKilled` (mirroring `AbstractBigquery`/`CustomJob`) and call `dataflow.projects().locations().jobs().update(..., JOB_STATE_CANCELLED or JOB_STATE_DRAINING)` from `kill()`, swallowing/logging any cancellation error rather than throwing.

**🟡 [Suggestion] `WaitForJob.java:100-118` — uses the deprecated `Await.until(Supplier, Duration, Duration)` overload**

`io.kestra.core.utils.Await.until(Supplier<T>, Duration, Duration)` is annotated `@Deprecated` in core in favor of `Await.await()`'s fluent Awaitility API. Not urgent, but worth migrating before core removes it.

**🟡 [Suggestion] Business requirement — sanity-check flow not included**

Linked issue #88's suggested task list item 10 calls for "Unit + integration tests and sanity-check YAML flow." Only mocked unit tests (`DataflowTest.java`) are present — no `src/test/resources/sanity-checks/*.yaml` + `RunnerTest`. Understandable given Dataflow can't run in Testcontainers/CI without real GCP credentials, but worth confirming with the issue owner whether a parse/validate-only sanity flow is still expected before closing #88.

**🟢 Nit** — `Trigger.lookback`'s implicit default (falls back to `interval` when unset, `Trigger.java:126`) isn't mentioned in its `@Schema(description = ...)` (`Trigger.java:103`). Worth a one-line addition so users know the default without reading the source.

**🟢 Nit** — Test task IDs in `DataflowTest.java` (`"launch-template"`, `"get-job"`, `"cancel-job"`, `"wait-for-job"`, `"trigger"`) are static rather than randomized, per the project's flaky-test-prevention convention. Low risk here since each test builds its own isolated task instance, but worth aligning.

---

### Summary

All 17 items from the two prior review rounds are genuinely fixed in the current diff — this is a clean, thorough follow-up pass, not a partial one. The remaining gap is the trigger's lack of persisted dedup state (real risk of duplicate/missed flow executions in production) and the missing `kill()` on `WaitForJob` (orphaned/billing Dataflow job on execution kill), both of which are concrete, fixable issues rather than nitpicks. Recommend addressing at least the trigger state-tracking issue before merge; the `kill()` gap can reasonably be a fast-follow if time-boxed, but should be tracked as a follow-up issue if not fixed here.

fdelbrayelle:
## Follow-up on previous review (re-review of `1f8cb41`, posted 2026-07-07)

Checked commit `31bdf35d` ("address review feedback, implement stateful trigger and kill hooks") against the two open findings from my last review.

### 1. `Trigger` — no persisted dedup state — **mostly addressed, one real gap**

`Trigger` now implements `StatefulTriggerInterface` and persists a state map via `StatefulTriggerService.readState`/`computeAndUpdateState`/`writeState` (`Trigger.java:158-214`), gated per candidate job (`id` as `uri`, `currentState` as `version`) instead of the old stateless `[now-lookback, now)` window scan. This is the right mechanism and matches `io.kestra.plugin.gcp.gcs.Trigger`'s pattern — verified `StatefulTriggerService`/`StatefulTriggerInterface` APIs exist as used (`core:1.3.19`).

However, the new `on` property is dead code:
- `Trigger.java:124-125` — `private final Property<On> on = Property.ofValue(On.CREATE_OR_UPDATE);` has no `@Schema`/`@PluginProperty(group = ...)`, unlike every other property in this class and unlike `gcs.Trigger`'s `on` field (`gcs/Trigger.java:126-139`).
- `Trigger.java:196` and `Trigger.java:213` both hardcode `On.CREATE_OR_UPDATE` directly instead of rendering `this.on` — compare with `gcs/Trigger.java:188-209`, which does `var rOn = runContext.render(on).as(On.class).orElse(On.CREATE_OR_UPDATE);` and passes `rOn` into `computeAndUpdateState`.

Net effect: the property is present in the builder/schema surface but a user setting `on: CREATE` or `on: UPDATE` gets `CREATE_OR_UPDATE` behavior regardless — silently ignored input. Please wire `runContext.render(this.on)` through at both call sites and add the missing `@Schema`/`@PluginProperty(group = ...)`.

Also note: no test exercises the dedup behavior itself (e.g. calling `evaluate()` twice for the same job/state and asserting the second call doesn't re-fire) — both `triggerEvaluation` tests are still single-call.

### 2. `WaitForJob` — missing `kill()`/`stop()` — **addressed**

`WaitForJob.java` now follows the `AbstractBigquery`/`CustomJob` pattern: `AtomicReference<Runnable> killable` + `AtomicBoolean isKilled` (both `@ToString.Exclude`), `killable` set once the job id is known, `kill()` guarded by `compareAndSet` and calling `dataflow...jobs().update(..., JOB_STATE_CANCELLED)` with the failure caught/logged rather than thrown, `stop()` delegating to `kill()`. The polling supplier also checks `isKilled` on every iteration. This correctly closes the orphaned-job/billing gap flagged previously. No test covers the new `kill()` path, but the implementation itself looks correct.

### Verdict: **Partially addressed**

The stateful-trigger and kill/stop mechanisms are both real, correctly-wired fixes for the two issues raised — but the `on` control on `Trigger` is non-functional and undocumented as shipped, which is a concrete, user-facing bug in the very code that was added to fix the dedup issue. Please fix the two call sites (`Trigger.java:196`, `Trigger.java:213`) to render `this.on`, and add `@Schema`/`@PluginProperty(group = ...)` to the field (`Trigger.java:124-125`), before merge.

fdelbrayelle:
Yet a little gap to fix @suraj-0628 please :pray:

fdelbrayelle (approved):

PR #655, opened 2026-07-02
fix(bigquery): avoid duplicate job submission on transient retry

### LINKED ISSUE #648: bigquery.CopyPartitions — retry submits new job without checking original job status, causing data duplication
## Summary

`io.kestra.plugin.gcp.bigquery.CopyPartitions` can produce duplicate data when BigQuery returns a transient `backendError` for a job that already completed successfully. The auto-retry mechanism submits a brand-new job without first verifying the original job's final status, so if both jobs succeed, data is copied twice.

## Actual Behaviour

When `waitForJob` receives a `BigQueryException` with reason `backendError` (or a message matching `"Retrying may solve the problem"`), it retries by calling `createJob.call()` again from scratch — allocating a new `JobId` and submitting a new BigQuery job. It does not query the original job's status before doing so. If BigQuery completed the original job successfully but returned the error transiently (a known API consistency quirk), the retry duplicates all data written by the first job.

## Expected Behaviour

Before submitting a new job on retry, the implementation should look up the original job's final status using the captured `JobId`. If the original job succeeded, the retry should be skipped and the successful result returned (or a clear warning raised), rather than silently submitting a duplicate operation.

## Reproducer

`CopyPartitions` task copying multiple source tables into a single destination table with `WRITE_APPEND`. One source table copy returns `backendError`; auto-retry fires and submits a second job; both jobs complete successfully; destination table contains duplicate rows.

```yaml
id: copy_partitions_reproducer
namespace: company.team

tasks:
  - id: copy
    type: io.kestra.plugin.gcp.bigquery.CopyPartitions
    sourceTables:
      - "project.dataset.source_table_*"
    destinationTable: "project.dataset.destination_table"
    writeDisposition: WRITE_APPEND
```

## Logs / Stack Trace

```
2026-06-30 11:05:45.303
Error query on job 'job_ShIcczOj2Thqk_PvnXKV4MMTGeWs' with errors:
[
 - BigQueryError{reason=backendError, location=null, message=Error encountered during execution. Retrying may solve the problem.}
]

2026-06-30 11:05:50.304
Retrying, attempts 1 elapsed 17 seconds

2026-06-30 11:05:56.192
Starting job 'JobId{project=gogame-pl, job=job_pgPvUG8vZfAxsXg-QB3k8yzzxkQL, location=US}'
```

Both jobs completed successfully on the BigQuery side, resulting in duplicated data in the destination table.

## Root Cause

In `AbstractBigquery.waitForJob`, the Failsafe retry policy wraps the entire `createJob.call()` invocation. On retry, `createJob` is called again from scratch — a new `JobId` is allocated and a new job is submitted. There is no pre-retry check to query whether the previous job (whose `JobId` was captured before the exception) completed successfully.

Relevant code path: `AbstractBigquery#waitForJob` → Failsafe retry → `createJob.call()` → new `Job` submitted.

## Environment

- **Plugin version**: 1.2.6 (reported); present in current `main`
- **Deployment**: Self-hosted

## Workaround

Set `retryReasons: []` and `retryMessages: []` on the task to disable auto-retry entirely. Transient `backendError`s will then fail the flow rather than self-healing, but data integrity is preserved.

```yaml
- id: copy
  type: io.kestra.plugin.gcp.bigquery.CopyPartitions
  retryReasons: []
  retryMessages: []
```

## Additional Context

The fix should query the original job's final status (via `bigquery.getJob(originalJobId)`) inside the retry lambda, before calling `createJob.call()` again. If the original job's status is `DONE` with no errors, return it directly without submitting a new job.

This pattern applies to all BigQuery tasks that go through `AbstractBigquery#waitForJob` (`Query`, `Load`, `LoadFromGcs`, `CopyPartitions`, `ExtractToGcs`, etc.).

---
*[View as Artifact](https://claude.ai/code/artifact/259ecc91-1b9b-43a2-bceb-a62b9806fe1e)*

fdelbrayelle:
## Summary

- `AbstractBigquery#waitForJob` re-submitted a brand-new BigQuery job on every Failsafe retry attempt, even when the previously submitted job had actually completed successfully but a transient error (e.g. `backendError`) occurred while polling/fetching its status — causing duplicate data (e.g. duplicated partitions in `CopyPartitions`).
- Before creating a new job on retry, the previous attempt's `JobId` is now tracked and looked up via `BigQuery#getJob(JobId)`. If that job is found and already `DONE` with no error, it is returned directly and `createJob.call()` is skipped for that attempt. If the job is still running, we block on `.waitFor()` to reach a terminal state before deciding (avoids racing two jobs). If the job is not found or done-with-error, behavior falls through unchanged to creating a new job.
- The first attempt (no prior `JobId`) is unaffected — same behavior as before.
- Updated the 4 callers of `waitForJob` (`Copy`, `Query`, `LoadFromGcs`, `Load`) to pass their existing `BigQuery connection`. `ExtractToGcs` does not go through `waitForJob` and was left untouched, as it is out of scope for this issue.
- Added a WireMock-based regression test in `QueryErrorTest` reproducing a transient polling failure against an already-succeeded job, asserting only one job creation call is made.

closes: https://github.com/kestra-io/plugin-gcp/issues/648

## Test plan

- [x] `rtk test ./gradlew test` passes for the bigquery package (`QueryErrorTest`, including the new `shouldNotDuplicateJobWhenPollingFailsAfterJobActuallySucceeded` test)
- [x] `./gradlew compileJava compileTestJava` builds cleanly
- [ ] Manual QA against a real BigQuery project with an injected transient `backendError` (not exercised in this session, no live GCP credentials available)

---
*[View as Artifact](https://claude.ai/code/artifact/e9815b2d-ff98-4589-ab67-0d8c98002389)*

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
@@ -102,7 +102,7 @@ abstract public class AbstractBigquery extends AbstractTask {
     }
 
     protected Job waitForJob(Logger logger, Callable<Job> createJob, Boolean dryRun, RunContext runContext, BigQuery connection) {
-        AtomicReference<JobId> lastJobId = new AtomicReference<>();
+        var lastJobId = new AtomicReference<JobId>();
 
         return Failsafe
             .with(
@@ -138,22 +138,27 @@ abstract public class AbstractBigquery extends AbstractTask {
             {
                 Job job = null;
                 try {
-                    JobId previousJobId = lastJobId.get();
-                    if (previousJobId != null) {
-                        Job previousJob = connection.getJob(previousJobId);
+                    // Dry-run jobs have no side effects and aren't reliably pollable, so always create a fresh one.
+                    if (!dryRun) {
+                        var previousJobId = lastJobId.get();
+                        if (previousJobId != null) {
+                            var previousJob = connection.getJob(previousJobId);
 
-                        if (previousJob != null) {
-                            if (!previousJob.isDone()) {
-                                previousJob = previousJob.waitFor();
-                            }
+                            if (previousJob != null) {
+                                if (!previousJob.isDone()) {
+                                    previousJob = previousJob.waitFor();
+                                }
 
-                            if (previousJob.getStatus().getError() == null) {
-                                logger.warn(
-                                    "Job '{}' already completed successfully despite a transient error, skipping duplicate retry",
-                                    previousJob.getJobId()
-                                );
+                                if (previousJob.getStatus().getError() == null) {
+                                    logger.warn(
+                                        "Job '{}' already completed successfully despite a transient error, skipping duplicate retry",
+                                        previousJob.getJobId()
+                                    );
 
-                                return previousJob;
+                                    return previousJob;
+                                }
+
+                                lastJobId.set(null);
                             }
                         }
                     }
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
@@ -6,6 +6,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicReference;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
 
@@ -96,11 +97,13 @@ abstract public class AbstractBigquery extends AbstractTask {
             .getService();
     }
 
-    protected Job waitForJob(Logger logger, Callable<Job> createJob, RunContext runContext) {
-        return this.waitForJob(logger, createJob, false, runContext);
+    protected Job waitForJob(Logger logger, Callable<Job> createJob, RunContext runContext, BigQuery connection) {
+        return this.waitForJob(logger, createJob, false, runContext, connection);
     }
 
-    protected Job waitForJob(Logger logger, Callable<Job> createJob, Boolean dryRun, RunContext runContext) {
+    protected Job waitForJob(Logger logger, Callable<Job> createJob, Boolean dryRun, RunContext runContext, BigQuery connection) {
+        AtomicReference<JobId> lastJobId = new AtomicReference<>();
+
         return Failsafe
             .with(
                 AbstractRetry.<Job> retryPolicy(
@@ -135,7 +138,28 @@ abstract public class AbstractBigquery extends AbstractTask {
             {
                 Job job = null;
                 try {
+                    JobId previousJobId = lastJobId.get();
+                    if (previousJobId != null) {
+                        Job previousJob = connection.getJob(previousJobId);
+
+                        if (previousJob != null) {
+                            if (!previousJob.isDone()) {
+                                previousJob = previousJob.waitFor();
+                            }
+
+                            if (previousJob.getStatus().getError() == null) {
+                                logger.warn(
+                                    "Job '{}' already completed successfully despite a transient error, skipping duplicate retry",
+                                    previousJob.getJobId()
+                                );
+
+                                return previousJob;
+                            }
+                        }
+                    }
+
                     job = createJob.call();
+                    lastJobId.set(job.getJobId());
 
                     BigQueryService.handleErrors(job, logger);
 
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/Load.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/Load.java
@@ -120,7 +120,7 @@ public class Load extends AbstractLoad implements RunnableTask<AbstractLoad.Outp
                     .build();
             }
 
-            Job job = this.waitForJob(logger, writer::getJob, runContext);
+            Job job = this.waitForJob(logger, writer::getJob, runContext, connection);
 
             return this.outputs(runContext, configuration, job);
         }
--- a/src/test/java/io/kestra/plugin/gcp/bigquery/QueryErrorTest.java
+++ b/src/test/java/io/kestra/plugin/gcp/bigquery/QueryErrorTest.java
@@ -334,7 +334,6 @@ public class QueryErrorTest {
         try {
             output = task.run(runContext);
         } catch (Exception e) {
-            getAllServeEvents().forEach(ev -> System.out.println("REQ: " + ev.getRequest().getMethod() + " " + ev.getRequest().getUrl() + " -> " + ev.getResponse().getStatus()));
             throw e;
         }
 
--- a/src/test/java/io/kestra/plugin/gcp/bigquery/QueryErrorTest.java
+++ b/src/test/java/io/kestra/plugin/gcp/bigquery/QueryErrorTest.java
@@ -2,6 +2,7 @@ package io.kestra.plugin.gcp.bigquery;
 
 import com.github.tomakehurst.wiremock.junit5.WireMockRuntimeInfo;
 import com.github.tomakehurst.wiremock.junit5.WireMockTest;
+import com.github.tomakehurst.wiremock.stubbing.Scenario;
 import com.google.cloud.NoCredentials;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryOptions;
@@ -85,7 +86,8 @@ public class QueryErrorTest {
           "status": { "state": "RUNNING" },
           "statistics": {
             "creationTime": "1697123456789",
-            "startTime": "1697123457000"
+            "startTime": "1697123457000",
+            "query": { "statementType": "SELECT" }
           }
         }""";
 
@@ -112,6 +114,70 @@ public class QueryErrorTest {
             "startTime": "1697123457000",
             "endTime": "1697123459000",
             "query": { "statementType": "SELECT" }
+          }
+        }""";
+
+    private static final String JOB_RESPONSE_DUP_TEST_DONE_WITH_ERROR = """
+        {
+          "kind": "bigquery#job",
+          "etag": "\\"abcdef1234567890\\"",
+          "id": "kestra-unit-test:europe-west3.job_dup_test",
+          "jobReference": {
+            "projectId": "kestra-unit-test",
+            "jobId": "job_dup_test",
+            "location": "europe-west3"
+          },
+          "configuration": {
+            "query": {
+              "query": "SELECT * FROM `kestra-unit-test.my_dataset.my_table`",
+              "useLegacySql": false
+            },
+            "jobType": "QUERY"
+          },
+          "status": {
+            "state": "DONE",
+            "errorResult": {
+              "reason": "backendError",
+              "message": "Backend error while processing the job"
+            }
+          },
+          "statistics": {
+            "creationTime": "1697123456789",
+            "startTime": "1697123457000",
+            "endTime": "1697123459000",
+            "query": { "statementType": "SELECT" }
+          }
+        }""";
+
+    // Job#waitFor() polls the /queries/{jobId} endpoint, which returns a GetQueryResultsResponse,
+    // not a Job resource: it needs "jobComplete" to be seen as done, unlike the /jobs/{jobId} shapes above.
+    private static final String QUERY_RESULTS_RESPONSE_COMPLETE = """
+        {
+          "kind": "bigquery#getQueryResultsResponse",
+          "etag": "\\"abcdef1234567890\\"",
+          "jobReference": {
+            "projectId": "kestra-unit-test",
+            "jobId": "job_dup_test",
+            "location": "europe-west3"
+          },
+          "totalRows": "0",
+          "jobComplete": true,
+          "cacheHit": false
+        }""";
+
+    private static final String JOB_NOT_FOUND_RESPONSE = """
+        {
+          "error": {
+            "code": 404,
+            "errors": [
+              {
+                "domain": "global",
+                "message": "Not found: Job kestra-unit-test:europe-west3.job_dup_test",
+                "reason": "notFound"
+              }
+            ],
+            "message": "Not found: Job kestra-unit-test:europe-west3.job_dup_test",
+            "status": "NOT_FOUND"
           }
         }""";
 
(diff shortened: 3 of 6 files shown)


fdelbrayelle (commented):
test rtk proxy raw call

src/test/java/io/kestra/plugin/gcp/bigquery/QueryErrorTest.java
```diff
@@ -128,7 +179,70 @@ void shouldRetryOnBackendErrorWhenFetchingResults(WireMockRuntimeInfo wmRuntimeI
...
-        verify(3, getRequestedFor(urlPathMatching(resultsPath)));
+        // job.waitFor() polls the same results endpoint and fails once before AbstractBigquery#waitForJob
+        // finds the already-completed job via getJob() and short-circuits without submitting a duplicate job;
+        // the 3 remaining hits come from the dedicated "fetch results" retry loop in Query#run.
+        verify(4, getRequestedFor(urlPathMatching(resultsPath)));
+        verify(1, postRequestedFor(urlPathMatching(jobsPath)));
+    }
+
+    @Test
+    void shouldNotDuplicateJobWhenPollingFailsAfterJobActuallySucceeded(WireMockRuntimeInfo wmRuntimeInfo) throws Exception {
```
fdelbrayelle, line 256: 🟠 [Guidelines] Missing test coverage for new retry branches
Problem: this new test only exercises the branch where the previous job is found and already DONE with no error. Three other branches introduced in AbstractBigquery#waitForJob have zero coverage: (1) previous job found but still RUNNING -> blocks on previousJob.waitFor(), (2) previous job found and DONE but WITH an error -> falls through to a new job, (3) connection.getJob(previousJobId) returns null (job not found) -> falls through to a new job. A regression in any of these (e.g. blocking forever on a stuck job, or failing to fall through on a genuinely failed previous job) would not be caught by CI today.
Fix: add WireMock scenarios for each branch, asserting the expected number of job-creation POSTs, mirroring the pattern used here for the already-succeeded case.

src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
```diff
@@ -96,11 +97,13 @@ protected static BigQuery connection(RunContext runContext, GoogleCredentials go
...
 
-    protected Job waitForJob(Logger logger, Callable<Job> createJob, RunContext runContext) {
-        return this.waitForJob(logger, createJob, false, runContext);
+    protected Job waitForJob(Logger logger, Callable<Job> createJob, RunContext runContext, BigQuery connection) {
+        return this.waitForJob(logger, createJob, false, runContext, connection);
     }
 
-    protected Job waitForJob(Logger logger, Callable<Job> createJob, Boolean dryRun, RunContext runContext) {
+    protected Job waitForJob(Logger logger, Callable<Job> createJob, Boolean dryRun, RunContext runContext, BigQuery connection) {
+        AtomicReference<JobId> lastJobId = new AtomicReference<>();
```
fdelbrayelle: 🟡 [Guidelines] Use var for new local variables
Problem: `AtomicReference<JobId> lastJobId`, and further down `JobId previousJobId` / `Job previousJob`, all declare explicit types where the right-hand side already makes the type unambiguous, violating the "use var for all local variables whenever legally possible" convention.
Fix: `var lastJobId = new AtomicReference<JobId>();`, `var previousJobId = lastJobId.get();`, `var previousJob = connection.getJob(previousJobId);`.

src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
```diff
@@ -135,7 +138,28 @@ AbstractRetry.<Job> retryPolicy(
             {
                 Job job = null;
                 try {
+                    JobId previousJobId = lastJobId.get();
+                    if (previousJobId != null) {
+                        Job previousJob = connection.getJob(previousJobId);
+
+                        if (previousJob != null) {
+                            if (!previousJob.isDone()) {
```
fdelbrayelle: 🟡 [Guidelines] Lookback ignores dryRun; duplicated success-check logic
Problem: this block runs unconditionally on every retry, even when the current call is a dry run (Query.dryRun=true). com.google.cloud.bigquery.Job#isDone()/#waitFor() throw UnsupportedOperationException (not a BigQueryException/JobException) when invoked on a job whose QueryJobConfiguration has dryRun=true — that would bypass shouldRetry's instanceof BigQueryException check and abort the task with a confusing, unrelated crash instead of surfacing the real transient error. In practice this is unreachable today because BigQuery never persists dry-run jobs (connection.getJob() returns null for a dry-run job id, so the branch is skipped) — but that is an undocumented external invariant with no test covering the dry-run + retry combination.
Separately, the getError() == null check below duplicates the success/failure definition already encoded in BigQueryService.handleErrors a few lines down; consider reusing that helper (try { handleErrors(previousJob, logger); return previousJob; } catch (BigQueryException e) { /* fall through */ }) so the two definitions cannot silently diverge.
Fix: short-circuit the lookback when dryRun is true, and add a regression test for dry-run + transient-retry.

src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
```diff
@@ -96,11 +97,13 @@ protected static BigQuery connection(RunContext runContext, GoogleCredentials go
             .getService();
     }
 
-    protected Job waitForJob(Logger logger, Callable<Job> createJob, RunContext runContext) {
-        return this.waitForJob(logger, createJob, false, runContext);
+    protected Job waitForJob(Logger logger, Callable<Job> createJob, RunContext runContext, BigQuery connection) {
```
fdelbrayelle, line 100: 🟡 [Guidelines] Backward compatibility of protected overloads
Problem: these two protected waitForJob overloads had their signature changed in place (added mandatory BigQuery connection param) rather than gaining a new overload. AbstractBigquery is a public abstract class; any external subclass calling the old 3-arg/4-arg signatures now fails to compile.
Fix: acceptable if the plugin module is versioned accordingly (protected API of an internal plugin class, low blast radius), but worth confirming this is an intentional/accepted break rather than an oversight.

src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
```diff
@@ -135,7 +138,28 @@ AbstractRetry.<Job> retryPolicy(
...
+                                    previousJob.getJobId()
+                                );
+
+                                return previousJob;
+                            }
+                        }
+                    }
+
                     job = createJob.call();
+                    lastJobId.set(job.getJobId());
```
fdelbrayelle, line 167: 🟡 [Performance] Stale lastJobId re-queried on repeated failures
Problem: lastJobId is only overwritten here, after createJob.call() succeeds. If a subsequent retry's createJob.call() itself throws before reaching this line, lastJobId still points at the earlier, already-confirmed-failed job, so every following retry pays for one extra connection.getJob(...) RPC re-verifying a job we already know failed, before falling through to create another one.
Fix: reset lastJobId (e.g. lastJobId.set(null)) on the done-with-error fallthrough path so subsequent retries skip straight to job creation.

src/main/java/io/kestra/plugin/gcp/bigquery/Load.java
```diff
@@ -120,7 +120,7 @@ public Output run(RunContext runContext) throws Exception {
                     .build();
             }
 
-            Job job = this.waitForJob(logger, writer::getJob, runContext);
+            Job job = this.waitForJob(logger, writer::getJob, runContext, connection);
```
fdelbrayelle, line 123: 🟡 [Performance] Extra RPC added to an already-idempotent call site
Problem: writer::getJob is a pure getter (TableDataWriteChannel#getJob() returns a cached field, no network call) that always returns the same Job reference once the writer is closed — retrying it never creates a new job, so Load was never exposed to the duplicate-submission bug this PR fixes. With this change, every Failsafe retry for Load now performs one extra connection.getJob(...) round-trip before falling back to the free writer.getJob() call, with no corresponding safety benefit.
Fix: no functional change needed (harmless), but a one-line comment here explaining the lookback is a no-op safety net for this call site would help a future reader.

fdelbrayelle (commented):
## Kestra Plugin Code Review

### Business Requirements — met
The fix targets the exact root cause described in #648: `AbstractBigquery#waitForJob` re-submitted a brand-new job on every Failsafe retry even when the previously-submitted job had already succeeded. Tracking `lastJobId` outside the lambda and checking `connection.getJob(previousJobId)` before calling `createJob.call()` again matches the issue's own suggested fix. `CopyPartitions` (the original reproducer) is covered transitively since it delegates to `Copy`. Excluding `ExtractToGcs` is correctly justified — it never goes through `waitForJob`/Failsafe at all (no retry policy wraps its job creation), so it was never exposed to this bug in the first place.

### Kestra Guidelines — 5 findings
- 🟠 Missing test coverage for 3 of the 4 new branches introduced in `AbstractBigquery#waitForJob` — only the *already-succeeded* short-circuit is tested (inline comment on the test file).
- 🟡 Several new local variables use explicit types where `var` is unambiguous (style guideline: "Use var for all local variables whenever legally possible").
- 🟡 The previous-job lookback runs unconditionally regardless of the *current* job's `dryRun` flag, and duplicates the "is this job successful" definition already encoded in `BigQueryService.handleErrors` (see inline comments).
- 🟡 `waitForJob`'s two `protected` overloads had their signatures changed in place (extra mandatory `connection` param) rather than adding a new overload — a compile-time break for any external subclass of the public `AbstractBigquery` class. Low blast radius given this is an internal plugin implementation detail, but worth a mention per "favor backward compatibility."
- 🟢 Nit: when the previous-job lookup itself throws (e.g. `connection.getJob()` hits a transient error), the existing catch block logs "Error query on create job" even though no new job creation was attempted that attempt — slightly misleading during debugging (not anchorable inline, that block is unchanged by this diff).

### Security (OWASP Top 10:2025 + KPS) — 0 findings
No secret handling, no new external I/O beyond the existing Google Cloud SDK client, no injection surface, no logging of sensitive data. `getJob`/`waitFor` calls reuse the same already-authenticated `BigQuery` connection scoped correctly per call site.

### Performance — 2 findings
- 🟡 On the "done with error" fallthrough, `lastJobId` is only overwritten once `createJob.call()` succeeds; if the new attempt itself throws transiently first, subsequent retries keep re-querying the same already-confirmed-failed previous job via an extra `getJob()` RPC before finally retrying creation.
- 🟡 `Load.java`'s `createJob` callable (`writer::getJob`) is a pure local getter that always returns the same cached `Job` — this call site was never vulnerable to duplicate submission, yet the new lookback now adds one extra network round-trip per retry there with no corresponding safety benefit.

**Verdict:** REQUEST CHANGES

_Note: an earlier stray review comment ("test rtk proxy raw call") was posted on this PR while debugging the review-posting tooling — please disregard it, it carries no review content._

src/test/java/io/kestra/plugin/gcp/bigquery/QueryErrorTest.java
```diff
@@ -128,7 +245,303 @@ void shouldRetryOnBackendErrorWhenFetchingResults(WireMockRuntimeInfo wmRuntimeI
...
+                .withBody(JOB_RESPONSE_RUNNING)));
+
+        Query task = buildQueryWithoutFetch(wmRuntimeInfo, 3);
+        RunContext runContext = TestsUtils.mockRunContext(runContextFactory, task, ImmutableMap.of());
+
+        Query.Output output;
+        try {
+            output = task.run(runContext);
+        } catch (Exception e) {
+            getAllServeEvents().forEach(ev -> System.out.println("REQ: " + ev.getRequest().getMethod() + " " + ev.getRequest().getUrl() + " -> " + ev.getResponse().getStatus()));
```
fdelbrayelle: 🟢 [Nit] Leftover debug println in test
Problem: this catch block dumps every WireMock serve event to stdout via System.out.println — harmless (only fires when the test fails) but reads as leftover troubleshooting code left in from developing the scenario-state stubs, inconsistent with the rest of the file.
Fix: remove the try/catch and debug dump, or replace with a proper assertion-failure message if the diagnostic is worth keeping.

fdelbrayelle (commented):
## Kestra Plugin Code Review (re-review of 2cb88af)

_Note: posted as COMMENT, not APPROVE — the GitHub API rejects self-approval since I'm authenticated as the PR author. Verdict below is APPROVE._

### Business Requirements — met
Fix still correctly targets #648's root cause; unchanged from prior pass.

### Kestra Guidelines — 0 blocking findings, 1 nit
All 4 requested fixes verified correct against the actual google-cloud-bigquery 2.59.0 bytecode (decompiled `Job#waitFor()`/`BigQueryImpl#getJob(JobId,...)`) and against CI's green "Java Tests Report" run:
- ✅ Test coverage gap closed: `shouldNotDuplicateJobWhenPreviousJobStillRunning`, `shouldCreateNewJobWhenPreviousJobDoneWithError`, `shouldCreateNewJobWhenPreviousJobNotFound` each correctly drive the branch they claim to (traced request counts by hand against the WireMock scenario state machines; they line up with the actual retry/reload call sequence the SDK performs for QUERY jobs).
- ✅ Dry-run guard: the whole lookback is now inside `if (!dryRun)`; `shouldSkipLookbackOnDryRunRetry` asserts zero GETs on both the job-status and polling paths, confirming `isDone()`/`waitFor()` are never invoked on a dry-run job.
- ✅ `var` applied to `lastJobId`, `previousJobId`, `previousJob`.
- ✅ `lastJobId.set(null)` now runs immediately on the done-with-error branch (before falling through to `createJob.call()`), which is a stronger fix than the minimal ask — it also prevents the stale-lookup cost on the *next* attempt even if that attempt's own `createJob.call()` throws before reaching the old reset point.
- 🟢 Nit: `QueryErrorTest.shouldNotDuplicateJobWhenPreviousJobStillRunning` (line ~336) left in a debug `System.out.println` dump of WireMock serve events in its catch block — harmless (only triggers on test failure) but looks like leftover troubleshooting scaffolding from writing the scenario-state stubs; worth removing for consistency with the rest of the file's clean style.

Confirmed no stale 3-arg/4-arg `waitForJob` call sites remain (all 4 production callers — `Copy`, `Query`, `LoadFromGcs`, `Load` — pass `connection` correctly), and `CopyPartitions` doesn't call `waitForJob` directly so it's unaffected.

The two previously-flagged suggestions that were intentionally left unchanged (protected overload signature break, `Load.java`'s harmless extra RPC) remain accepted trade-offs per the prior instruction — not re-raised.

### Security (OWASP Top 10:2025 + KPS) — 0 findings
No change from prior pass: no secret handling, no new external I/O surface, reuses the already-authenticated `BigQuery connection` for both `getJob` and job creation.

### Performance — 0 findings
The stale-`lastJobId`-requery issue is now fixed at its root (reset happens before fallthrough, not after a successful subsequent creation), which is better than what was asked for.

**Verdict:** APPROVE

src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
fdelbrayelle: Kept as-is — the approved implementation plan already considered this tradeoff explicitly and accepted it (protected API on an internal plugin class, low blast radius). Not changing.

src/main/java/io/kestra/plugin/gcp/bigquery/Load.java
fdelbrayelle: Kept as-is per the same plan rationale — applying the lookback uniformly across all 4 call sites was an explicit design choice for consistency, accepted as costing nothing meaningful even where (as here) it's a no-op safety net.

Malaydewangan09 (approved):

PR #657, opened 2026-07-03
fix(bigquery): cancel underlying BigQuery job on task kill

### LINKED ISSUE #656: BigQuery job tasks don't cancel the BQ job on flow kill (e.g. SLA)
## Summary

BigQuery job-submitting tasks (`io.kestra.plugin.gcp.bigquery.Query` and the other tasks extending `AbstractJob`: `Load`, `Copy`, `CopyPartitions`, `ExtractToGcs`, `StorageWrite`) do not implement task lifecycle hooks to cancel the underlying BigQuery job when the Kestra execution is killed. When a flow is killed (e.g. by an SLA violation), the BigQuery job keeps running server-side, continuing to consume slots and billing after the flow has stopped.

## Actual Behaviour

None of the BigQuery job-submitting tasks override `stop()`/`kill()`. Killing a flow (manually or via SLA) terminates the Kestra execution but leaves the associated BigQuery job (`jobId`) in `RUNNING` state on the BigQuery side. Nothing calls the BigQuery `Jobs.cancel` API, so the job runs to completion (or keeps consuming resources) with no way for Kestra to stop it.

## Expected Behaviour

When a task is killed, Kestra should cancel the in-flight BigQuery job via the `Jobs.cancel` API before tearing down the task, the same way `io.kestra.plugin.gcp.vertexai.CustomJob` already implements `kill()` to cancel its in-flight Vertex AI job (see `src/main/java/io/kestra/plugin/gcp/vertexai/CustomJob.java`).

## Reproducer

No reproducer flow needed — the gap is structural (missing lifecycle hook), not input-dependent. Steps to observe:

```yaml
id: reproducer
namespace: company.team

tasks:
  - id: long_running_query
    sql: "SELECT * FROM \`bigquery-public-data.samples.natality\` LIMIT 1000000"

sla:
  - type: MAX_DURATION
    duration: PT5S
    behavior: KILL
```

Trigger the SLA (or manually kill the execution) while the query job is running, then check the BigQuery Jobs console/API: the job is still `RUNNING` or completes normally, even though the Kestra execution shows `KILLED`.

## Logs / Stack Trace

No logs available — this is a missing-behavior gap, not a runtime error.

## Environment

- **Kestra version**: 
- **Plugin version**: 
- **Deployment**: 

## Additional Context

Surfaced from the same investigation as a separate BigQuery job-duplication issue on `CopyPartitions` (retry submitting a second job without checking the first job's final state), which was fixed in `v2.7.3`. This one is a distinct gap: no job cancellation on kill.

Implementation pointer: `RunnableTask` implementations can hook into `stop()` (graceful shutdown) or `kill()` (immediate shutdown) to run cleanup logic. `io.kestra.plugin.gcp.vertexai.CustomJob` already implements this pattern (`kill()` calls `JobServiceClient#cancelCustomJob`); the BigQuery job tasks should do the equivalent with the BigQuery `Jobs.cancel` API.

---
*[View as Artifact](https://claude.ai/code/artifact/38c553d9-3863-4ec3-b09f-b86f66841761)*

fdelbrayelle:
## Summary

- `AbstractBigquery` now tracks the `BigQuery` connection and `JobId` of the job currently in flight (`trackJob()`) and overrides `kill()` and `stop()` to call `BigQuery#cancel(JobId)` on it, guarded by a shared `AtomicBoolean` so a duplicate/racing kill and/or stop is a no-op regardless of which lifecycle event fires first.
- `waitForJob()` calls `trackJob()` right after every `createJob.call()`, including retries, so a kill or stop during a retry cancels the live job rather than a stale one. `Query`, `Copy`, `Load`, and `LoadFromGcs` all funnel through it and get the fix for free.
- `ExtractToGcs` tracks its extract job directly since it doesn't go through `waitForJob()`.
- `CopyPartitions` builds a nested `Copy` task per run; it now stores a reference to it and forwards both `kill()` and `stop()` to it, same as `kill()`.
- `StorageWrite` and `DeletePartitions` are intentionally untouched — neither submits a cancellable Jobs-API job.

closes: https://github.com/kestra-io/plugin-gcp/issues/656

## Test plan

- [x] `rtk test ./gradlew test` passes (full suite, 0 failures/errors)
- [ ] `./gradlew shadowJar` builds
- [x] New unit tests: `AbstractBigqueryLifecycleTest` (cancel called once via `kill()` or `stop()`, no-op on second kill/stop or kill-after-stop/stop-after-kill, no-op with no tracked job, retry tracks latest job only, cancel exceptions swallowed) and `CopyPartitionsKillTest` (forwarding of both `kill()` and `stop()` to the nested `Copy` task)

---
*[View as Artifact](https://claude.ai/code/artifact/a3ae95f2-b933-4625-8fb6-e84f9813a623)*

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
@@ -85,16 +85,29 @@ abstract public class AbstractBigquery extends AbstractTask implements WorkerJob
 
     @JsonIgnore
     @Getter(AccessLevel.NONE)
+    @EqualsAndHashCode.Exclude
+    @ToString.Exclude
     @Builder.Default
     private final AtomicReference<BigQuery> trackedConnection = new AtomicReference<>();
 
     @JsonIgnore
     @Getter(AccessLevel.NONE)
+    @EqualsAndHashCode.Exclude
+    @ToString.Exclude
     @Builder.Default
     private final AtomicReference<JobId> trackedJobId = new AtomicReference<>();
 
     @JsonIgnore
     @Getter(AccessLevel.NONE)
+    @EqualsAndHashCode.Exclude
+    @ToString.Exclude
+    @Builder.Default
+    private final AtomicReference<Logger> trackedLogger = new AtomicReference<>();
+
+    @JsonIgnore
+    @Getter(AccessLevel.NONE)
+    @EqualsAndHashCode.Exclude
+    @ToString.Exclude
     @Builder.Default
     private final AtomicBoolean isCancelled = new AtomicBoolean(false);
 
@@ -102,9 +115,10 @@ abstract public class AbstractBigquery extends AbstractTask implements WorkerJob
      * Records the job currently submitted, so that {@link #kill()} or {@link #stop()} can cancel the
      * live BigQuery job instead of a stale one from a previous retry attempt.
      */
-    protected void trackJob(BigQuery connection, JobId jobId) {
+    protected void trackJob(BigQuery connection, JobId jobId, Logger logger) {
         this.trackedConnection.set(connection);
         this.trackedJobId.set(jobId);
+        this.trackedLogger.set(logger);
     }
 
     @Override
@@ -126,7 +140,12 @@ abstract public class AbstractBigquery extends AbstractTask implements WorkerJob
                 try {
                     connection.cancel(jobId);
                 } catch (Exception e) {
-                    LOG.warn("Failed to cancel BigQuery job '{}'", jobId, e);
+                    Logger logger = this.trackedLogger.get();
+                    if (logger != null) {
+                        logger.warn("Failed to cancel BigQuery job '{}'", jobId, e);
+                    } else {
+                        LOG.warn("Failed to cancel BigQuery job '{}'", jobId, e);
+                    }
                 }
             }
         }
@@ -219,7 +238,7 @@ abstract public class AbstractBigquery extends AbstractTask implements WorkerJob
 
                     job = createJob.call();
                     lastJobId.set(job.getJobId());
-                    this.trackJob(connection, job.getJobId());
+                    this.trackJob(connection, job.getJobId(), logger);
 
                     BigQueryService.handleErrors(job, logger);
 
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
@@ -96,11 +96,11 @@ abstract public class AbstractBigquery extends AbstractTask implements WorkerJob
     @JsonIgnore
     @Getter(AccessLevel.NONE)
     @Builder.Default
-    private final AtomicBoolean isKilled = new AtomicBoolean(false);
+    private final AtomicBoolean isCancelled = new AtomicBoolean(false);
 
     /**
-     * Records the job currently submitted, so that {@link #kill()} can cancel the live BigQuery job
-     * instead of a stale one from a previous retry attempt.
+     * Records the job currently submitted, so that {@link #kill()} or {@link #stop()} can cancel the
+     * live BigQuery job instead of a stale one from a previous retry attempt.
      */
     protected void trackJob(BigQuery connection, JobId jobId) {
         this.trackedConnection.set(connection);
@@ -109,7 +109,16 @@ abstract public class AbstractBigquery extends AbstractTask implements WorkerJob
 
     @Override
     public void kill() {
-        if (isKilled.compareAndSet(false, true)) {
+        cancelTrackedJob();
+    }
+
+    @Override
+    public void stop() {
+        cancelTrackedJob();
+    }
+
+    private void cancelTrackedJob() {
+        if (isCancelled.compareAndSet(false, true)) {
             BigQuery connection = this.trackedConnection.get();
             JobId jobId = this.trackedJobId.get();
 
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
@@ -6,16 +6,19 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 
 
 import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.auth.oauth2.GoogleCredentials;
 import com.google.cloud.bigquery.*;
 
 import io.kestra.core.exceptions.IllegalVariableEvaluationException;
+import io.kestra.core.models.WorkerJobLifecycle;
 import io.kestra.core.models.annotations.PluginProperty;
 import io.kestra.core.models.property.Property;
 import io.kestra.core.models.tasks.retrys.AbstractRetry;
@@ -33,7 +36,9 @@ import lombok.experimental.SuperBuilder;
 @EqualsAndHashCode
 @Getter
 @NoArgsConstructor
-abstract public class AbstractBigquery extends AbstractTask {
+abstract public class AbstractBigquery extends AbstractTask implements WorkerJobLifecycle {
+    private static final Logger LOG = LoggerFactory.getLogger(AbstractBigquery.class);
+
     @Schema(
         title = "Dataset location",
         description = "Optional BigQuery location for created or targeted resources. Experimental and may change; see BigQuery dataset location documentation."
@@ -77,6 +82,46 @@ abstract public class AbstractBigquery extends AbstractTask {
             "Retrying may solve the problem"
         )
     );
+
+    @JsonIgnore
+    @Getter(AccessLevel.NONE)
+    @Builder.Default
+    private final AtomicReference<BigQuery> trackedConnection = new AtomicReference<>();
+
+    @JsonIgnore
+    @Getter(AccessLevel.NONE)
+    @Builder.Default
+    private final AtomicReference<JobId> trackedJobId = new AtomicReference<>();
+
+    @JsonIgnore
+    @Getter(AccessLevel.NONE)
+    @Builder.Default
+    private final AtomicBoolean isKilled = new AtomicBoolean(false);
+
+    /**
+     * Records the job currently submitted, so that {@link #kill()} can cancel the live BigQuery job
+     * instead of a stale one from a previous retry attempt.
+     */
+    protected void trackJob(BigQuery connection, JobId jobId) {
+        this.trackedConnection.set(connection);
+        this.trackedJobId.set(jobId);
+    }
+
+    @Override
+    public void kill() {
+        if (isKilled.compareAndSet(false, true)) {
+            BigQuery connection = this.trackedConnection.get();
+            JobId jobId = this.trackedJobId.get();
+
+            if (connection != null && jobId != null) {
+                try {
+                    connection.cancel(jobId);
+                } catch (Exception e) {
+                    LOG.warn("Failed to cancel BigQuery job '{}'", jobId, e);
+                }
+            }
+        }
+    }
 
     BigQuery connection(RunContext runContext) throws IllegalVariableEvaluationException, IOException {
         GoogleCredentials credentials = this.credentials(runContext);
@@ -165,6 +210,7 @@ abstract public class AbstractBigquery extends AbstractTask {
 
                     job = createJob.call();
                     lastJobId.set(job.getJobId());
+                    this.trackJob(connection, job.getJobId());
 
                     BigQueryService.handleErrors(job, logger);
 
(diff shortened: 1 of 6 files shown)


src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
```diff
@@ -78,6 +83,55 @@ abstract public class AbstractBigquery extends AbstractTask {
...
+    private void cancelTrackedJob() {
+        if (isCancelled.compareAndSet(false, true)) {
+            BigQuery connection = this.trackedConnection.get();
+            JobId jobId = this.trackedJobId.get();
+
+            if (connection != null && jobId != null) {
+                try {
+                    connection.cancel(jobId);
+                } catch (Exception e) {
+                    LOG.warn("Failed to cancel BigQuery job '{}'", jobId, e);
```
fdelbrayelle: 🟠 [Guidelines] Static Logger instead of `runContext.logger()` in kill/stop path
Problem: `cancelTrackedJob()` logs a cancel failure via a static SLF4J `Logger`, so it only ever reaches worker/server logs — never the Kestra execution's log console the flow author actually looks at. The guidelines mandate `runContext.logger()` for task logging, and the reference implementation this fix is modeled on (`CustomJob.safelyKillJob`) deliberately captures `runContext` in its cancellation closure for exactly this reason.
Fix: Thread `RunContext` through `trackJob(...)` — it's already in scope at both call sites (`waitForJob(..., RunContext runContext, ...)` and `ExtractToGcs.run(RunContext runContext)`) — and use `runContext.logger()` in `cancelTrackedJob()` instead of the static `LOG`.

src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
```diff
@@ -78,6 +83,55 @@ abstract public class AbstractBigquery extends AbstractTask {
         )
     );
 
+    @JsonIgnore
+    @Getter(AccessLevel.NONE)
+    @Builder.Default
+    private final AtomicReference<BigQuery> trackedConnection = new AtomicReference<>();
```
fdelbrayelle, line 91: 🟠 [Guidelines] Internal tracking fields not excluded from generated equals/hashCode/toString
Problem: `trackedConnection`, `trackedJobId` and `isCancelled` get `@JsonIgnore` + `@Getter(AccessLevel.NONE)` but not `@EqualsAndHashCode.Exclude`/`@ToString.Exclude`. `AtomicReference`/`AtomicBoolean` don't override `equals()`, so with the class-level `@EqualsAndHashCode` already mandatory on this class, two structurally-identical `Query`/`Copy`/`Load`/`LoadFromGcs`/`ExtractToGcs` instances built separately will never be `.equals()` to each other (reference-identity comparison on these fields), and both `hashCode()` and `toString()` now mutate over the task's lifetime as a job gets tracked. This affects every `AbstractBigquery` subclass, not just the ones this PR touches.
Fix: Add `@EqualsAndHashCode.Exclude` and `@ToString.Exclude` on all three fields (same treatment `CopyPartitions.copyTask` needs, see other comment).

src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
```diff
@@ -78,6 +83,55 @@ abstract public class AbstractBigquery extends AbstractTask {
...
+    @JsonIgnore
+    @Getter(AccessLevel.NONE)
+    @Builder.Default
+    private final AtomicBoolean isCancelled = new AtomicBoolean(false);
+
+    /**
+     * Records the job currently submitted, so that {@link #kill()} or {@link #stop()} can cancel the
+     * live BigQuery job instead of a stale one from a previous retry attempt.
+     */
+    protected void trackJob(BigQuery connection, JobId jobId) {
```
fdelbrayelle: 🟡 [Performance] Two narrow races around the tracked-job guard
Problem: (1) `trackJob()` writes `trackedConnection` then `trackedJobId` as two separate non-atomic operations; a `cancelTrackedJob()` racing in between could read a stale `jobId` (low impact here since `connection` is invariant across retries within one `waitForJob()` call). (2) If `kill()`/`stop()` fires before the very first `trackJob()` call (e.g. between task dispatch and `createJob.call()`), `isCancelled.compareAndSet(false, true)` consumes the one-shot guard on a no-op — the job submitted moments later is then never cancelled, since the guard is spent. This mirrors the existing `CustomJob.killable`/`isKilled` pattern already accepted in this codebase, so not a new class of bug, but it now has a wider blast radius (every `AbstractBigquery` subclass) than a single task.
Fix: not blocking given precedent; consider a single atomically-set holder for connection+jobId, and/or re-checking `isCancelled` at the end of `trackJob()` to catch a signal that arrived too early.

src/main/java/io/kestra/plugin/gcp/bigquery/CopyPartitions.java
```diff
@@ -110,6 +112,11 @@ public class CopyPartitions extends AbstractPartition implements RunnableTask<Co
     )
     protected Property<Boolean> skipEmpty = Property.ofValue(false);
 
+    @JsonIgnore
+    @Getter(AccessLevel.NONE)
+    @Builder.Default
+    private final AtomicReference<Copy> copyTask = new AtomicReference<>();
```
fdelbrayelle, line 120: 🟠 [Guidelines] Same equals/hashCode/toString exposure as AbstractBigquery's tracking fields
Problem: `copyTask` is `@JsonIgnore` + `@Getter(AccessLevel.NONE)` but not excluded from `@EqualsAndHashCode`/`@ToString`, so `CopyPartitions` instances are subject to the same reference-identity equals()/mutable-hashCode issue described on `AbstractBigquery.trackedConnection`.
Fix: Add `@EqualsAndHashCode.Exclude` and `@ToString.Exclude` here as well.

fdelbrayelle (commented):
## Kestra Plugin Code Review

### Business Requirements — met
Addresses issue #656 exactly: all BigQuery tasks that actually submit a cancellable Jobs-API job (`Query`, `Copy`, `Load`, `LoadFromGcs`, `ExtractToGcs`, and `CopyPartitions` via its nested `Copy`) now cancel that job on `kill()`, and additionally on `stop()` (graceful shutdown), which is a sound extension beyond the literal issue text. `StorageWrite` (gRPC streaming, no Jobs-API job) and `DeletePartitions` (synchronous REST partition deletes, no Job at all) are correctly and verifiably excluded — confirmed neither calls `waitForJob()`/creates a `Job`. Shared logic is centralized in `AbstractBigquery` rather than duplicated per task, matching the guideline's 'prefer a shared base class' recommendation and the `CustomJob` reference pattern the issue points to. Scope is clean — only the 5 claimed files are touched, no unrelated changes.

### Kestra Guidelines — 3 findings
🟠 Static `Logger` used instead of `runContext.logger()` in the kill/stop cancellation path (`AbstractBigquery.cancelTrackedJob()`), so a failed cancel only surfaces in worker/server logs, not the execution's log console — see inline comment.
🟠 New internal tracking fields (`AbstractBigquery.trackedConnection`/`trackedJobId`/`isCancelled`, `CopyPartitions.copyTask`) are hidden from the getter/serialization surface but not excluded from the mandatory `@EqualsAndHashCode`/`@ToString` — see inline comments (2 locations).

### Security (OWASP Top 10:2025 + KPS) — ✅ No issues found
No secrets handled in this diff, cancellation failures are caught and logged (never rethrown, never blocking `kill()`/`stop()`), no injection surface, no new dependencies, no leaked internals in error messages (`JobId` isn't sensitive).

### Performance — 1 finding
🟡 Two narrow lifecycle-hook race windows in `AbstractBigquery.trackJob()`/`cancelTrackedJob()` (non-atomic two-field write; one-shot cancel guard that can be consumed by a kill/stop arriving before the first job is tracked). Both mirror the existing `CustomJob.killable`/`isKilled` pattern already accepted in this codebase, so not a novel regression, but the blast radius is wider here since it's now in a shared base class used by ~8 task types. Non-blocking — see inline comment for detail.

**Verdict:** REQUEST CHANGES (posted as COMMENT — GitHub disallows REQUEST_CHANGES/APPROVE on one's own pull request)

fdelbrayelle (commented):
## Kestra Plugin Code Review (re-review after fix batch 700c237)

### Fix verification — both prior findings resolved
1. **Logger visibility (was 🟠)** — Fixed correctly. `trackJob(BigQuery, JobId, Logger)` now takes and stores the execution logger in a new `AtomicReference<Logger> trackedLogger`. Both call sites pass the in-scope `runContext.logger()`: `waitForJob()` at `AbstractBigquery.java:241` (`this.trackJob(connection, job.getJobId(), logger)`, where `logger` is the method parameter fed by `Query`/`Copy`/`Load`/`LoadFromGcs` from `runContext.logger()`), and `ExtractToGcs.run()` at line 135. `cancelTrackedJob()` now logs via `trackedLogger.get()` with a correct null-safe fallback to the static `LOG` only when no logger was ever tracked.
2. **equals/hashCode/toString exclusion (was 🟠 ×2)** — Fixed correctly. All four `AbstractBigquery` tracking fields (`trackedConnection`, `trackedJobId`, `trackedLogger`, `isCancelled`) and `CopyPartitions.copyTask` now carry `@EqualsAndHashCode.Exclude` + `@ToString.Exclude` alongside the existing `@JsonIgnore`/`@Getter(AccessLevel.NONE)`. No field was missed.

### Re-verification checklist
- Null-safety of `trackedLogger` when `cancelTrackedJob()` runs before any `trackJob()` call: safe — `trackedConnection`/`trackedJobId` are also null in that case, so the `connection != null && jobId != null` guard short-circuits before the logger is ever touched. Covered by `killWithoutTrackedJobIsNoOp`/`stopWithoutTrackedJobIsNoOp`.
- `AbstractBigqueryLifecycleTest` and `CopyPartitionsKillTest`: all call sites consistently use the new 3-arg `trackJob(connection, jobId, LOGGER)` signature; no stale 2-arg calls remain.
- Scope discipline: `git diff main...HEAD --name-only` shows exactly the 5 expected files (`AbstractBigquery.java`, `CopyPartitions.java`, `ExtractToGcs.java`, `AbstractBigqueryLifecycleTest.java`, `CopyPartitionsKillTest.java`) — no drift.
- Third finding (🟡 non-blocking race window on `trackJob()`'s sequential field writes + one-shot `isCancelled` guard) — confirmed untouched by this batch, as instructed. The window now spans three sequential writes instead of two (connection → jobId → logger), but the added third write only affects which logger a rare mis-timed cancel uses, not whether cancellation itself fires correctly (connection+jobId are still written first). Not re-flagged as blocking, same acceptance rationale as before (mirrors the accepted `CustomJob` pattern).
- `isKilled` → `isCancelled` rename (to reflect it now guards both `kill()` and `stop()`) is clean; no leftover references outside the unrelated `CustomJob` reference implementation.
- `CopyPartitions.kill()`/`stop()` correctly `@Override` `AbstractBigquery`'s implementation (via `AbstractPartition → AbstractTable → AbstractBigquery`) to delegate to the nested `Copy` task instead of the (never-populated) self-tracked job — correct design, since `CopyPartitions.run()` never calls `trackJob()` itself.
- No new dependencies introduced; `mockito-junit-jupiter` was already a `testImplementation`.

### Kestra Guidelines — 0 blocking findings
🟢 Nit (non-blocking, pre-existing pattern, not part of this fix batch): `cancelTrackedJob()`'s three local reads (`BigQuery connection = ...`, `JobId jobId = ...`, `Logger logger = ...`) use explicit types where `var` is legal and would be consistent with the "use `var` for all local variables" guideline. This predates the current fix batch (present since the original `kill()` implementation) and wasn't flagged in the previous review round; calling it out only for completeness, not blocking.

### Security (OWASP Top 10:2025 + KPS) — ✅ No issues found
No change to the security posture from the previous pass: cancellation failures are still caught and logged, never rethrown; no secrets involved; no new external calls or dependencies.

### Performance — ✅ No new issues found
The only performance-track item from the previous round (narrow race window) was intentionally left as-is per instruction and remains non-blocking for the reasons stated above.

**Verdict:** APPROVE

src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
fdelbrayelle: Fixed in 700c237 — `trackJob()` now takes a `Logger` param (the execution logger from `runContext.logger()`), stores it in `trackedLogger`, and `cancelTrackedJob()` logs the cancel-failure warning through it instead of the static `LOG`, matching the `CustomJob#kill()` pattern.

src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
fdelbrayelle: Fixed in 700c237 — added `@EqualsAndHashCode.Exclude` and `@ToString.Exclude` on `trackedConnection`, `trackedJobId`, `isCancelled`, and `trackedLogger` so the mutable tracking state no longer affects equality semantics or `toString()` output over the task instance's lifecycle.

src/main/java/io/kestra/plugin/gcp/bigquery/AbstractBigquery.java
fdelbrayelle: Left as-is per your own non-blocking assessment. The two-field write in `trackJob()` and the one-shot cancel guard mirror the already-accepted `CustomJob` `killable`/`isKilled` pattern used elsewhere in this plugin, and the re-review confirmed this should stay untouched.

src/main/java/io/kestra/plugin/gcp/bigquery/CopyPartitions.java
fdelbrayelle: Fixed in 700c237 — same `@EqualsAndHashCode.Exclude` / `@ToString.Exclude` treatment applied to `copyTask` here.

jymaire (approved):

PR #658, opened 2026-07-03
feat(topology): resolve Pebble expressions in BigQuery Query topology view

jymaire:
## Summary

Ports [plugin-ee-aws#170](https://github.com/kestra-io/plugin-ee-aws/pull/170) to plugin-gcp.

  - BigqueryQueryTopologyDetails resolves Pebble expressions in projectId, location, and sql for display via  renderExpressions() (POST /expressions/render), inline in the component — best-effort, display-only, raw value  kept on failure.
  - Active tenant read from persisted localStorage (selectedTenant) so requests hit the right tenant on non-main EE;  "main" default is correct on OSS.
  - Resolution context: executionId → namespace + flowId → globals.
  - Bumps @kestra-io/kestra-sdk to a 2.0.0 prerelease exposing the ./expressions subpath.
  - Documents the yalc local-linking workflow (ui/README.md); gitignores .yalc/yalc.lock.

## Depends on

> ⚠️ Needs a published `@kestra-io/kestra-sdk` that exposes the `./expressions` subpath (the `renderExpressions()` export). Same SDK dependency as the AWS PR.

## Test plan
- [ ] `{{ vars.projectId }}` / `{{ vars.location }}` defined as namespace variables resolve in the Show Details / topology view
- [ ] Unresolvable expressions fall back to the raw value (no crash)
- [ ] Resolves against the correct tenant on a non-`main` EE instance


jymaire:
# QA

## Flow

```yaml
id: test-bq-topology-expressions
namespace: company.team

  # Flow-level variables holding the values the task references through Pebble.
  # The topology "Show Details" panel should display the RESOLVED values
  # (e.g. "my-gcp-project", "EU") instead of the raw "{{ vars.projectId }}".
variables:
    projectId: my-gcp-project
    location: EU

inputs:
    # Alternative to vars: flip the task to use {{ inputs.* }} to test input resolution.
    - id: projectId
      type: STRING
      defaults: my-gcp-project
    - id: location
      type: STRING
      defaults: EU

tasks:
    - id: query
      projectId: "{{ vars.projectId }}"
      location: "{{ vars.location }}"
      sql: |
        SELECT
          "{{ vars.projectId }}" AS project,
          "{{ vars.location }}" AS location,
          CURRENT_TIMESTAMP() AS ran_at


```

## Screen

image

jymaire:
## Why the plugin renders its own resolved SQL block instead of fixing the host's source preview

A reasonable question on this PR is: the topology panel already shows the task's SQL in the host's built-in source preview — why does the plugin render a *second*, resolved copy underneath instead of just making the existing preview show resolved values? Short version: the plugin can resolve the value, but it can't own the surface the host preview is painted into.

### Resolution is not the differentiator

`projectId`, `location`, **and** `sql` all go through the same path — one `POST /expressions/render` call:

```ts
useRenderedExpressions(
    () => [projectId.value, location.value, sql.value],  // all three resolved together
    ...
)
```

The backend happily returns the resolved SQL. So "who resolves the expression" is not the issue — the difference is **who owns the DOM surface the text is displayed in.**

### Two kinds of surface in the detail panel

```
┌─ Topology detail panel — a HOST Vue component tree ─────────────┐
│                                                                 │
│  ┌─ Task-source preview  (HOST renders this, via Monaco) ──┐    │  host owns this DOM,
│  │  sql: |                                                 │    │  fed from raw flow YAML
│  │    SELECT "{{ vars.projectId }}" AS project ...         │    │  → shows raw {{ }}
│  └─────────────────────────────────────────────────────────┘    │
│                                                                 │
│  ┌─ slot "topology-details"  (PLUGIN renders this) ────────┐    │  we own this DOM;
│  │   Project    my-gcp-project          ← resolved         │    │  we choose what
│  │   Location   EU                      ← resolved         │    │  goes in it
│  │   Query   ┌────────────────────────────────────┐       │    │
│  │           │ SELECT "my-gcp-project" AS ...      │  ← resolved copy the plugin adds
│  │           └────────────────────────────────────┘       │    │
│  └─────────────────────────────────────────────────────────┘    │
└─────────────────────────────────────────────────────────────────┘
```

- **Project / Location / the resolved Query block** live inside the `topology-details` **slot** — the mount point the host explicitly hands the plugin. We create those elements, so we decide what goes in them, and we put the resolved values there.
- **The host source preview** is Kestra core's built-in task-source view. The **host's** Vue component created that Monaco editor and feeds it from the **host's** own reactive state (the raw flow YAML). The plugin was never handed a reference to it.

### Why we can't just reach in and edit the host preview

The plugin shares the same `document` (module federation, single page), so in principle `document.querySelector(...)` could find the Monaco node and overwrite its text. It fails for three compounding reasons:

1. **Vue owns that subtree.** On the host's next re-render (theme toggle, resize, tab switch, refetch), Vue reconciles from its virtual DOM and clobbers our mutation.
2. **Monaco is model-driven.** Its content is bound to host props; even setting text imperatively gets reverted the moment the host pushes `model-value` again, and can desync editor state in between.
3. **No stability contract.** That block's DOM structure and class names are host internals — a Kestra core version bump can restructure them silently and break the selector with no compile-time warning.

The `@kestra-io/artifact-sdk` exposes exactly **two** plugin mount points (`topology-details`, `topology-task-drawer`) and **no** source/code slot. Those two slots are the only surfaces where the host guarantees the plugin's output survives its re-renders.

### The rule, and the tradeoff

> The plugin can resolve *any* property value (the backend renders them all), but it can only **display** that value inside a surface the host hands it. The source preview isn't such a surface — it's host chrome painted from raw YAML by host code the plugin can't drive.

Hence this PR's pragmatic choice: rather than fight the host preview, the plugin renders its own resolved Query block in its slot (accepting that the raw host preview and the resolved plugin copy coexist).

The only clean way to make the **host preview itself** show resolved SQL is a **Kestra core change** — either resolve expressions in the built-in preview, or expose a slot/prop so a plugin can feed it. Happy to open a separate core issue/PR for that if we want to eliminate the duplication.

fdelbrayelle (approved):

PR #666, opened 2026-07-20
fix: resolve GCloudCLI credential path via task runner working dir

Skraye:
GCloudCLI writes the service account key to a temp file in the working dir but exposed the worker-side absolute path in GOOGLE_APPLICATION_CREDENTIALS / CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE. 
That path only resolves when the runner bind-mounts the working dir at the same path (Docker). On the Kubernetes runner the working dir is remapped to /kestra/working-dir, so gcloud couldn't find the key and authentication failed.

The path now goes through taskRunner.toAbsolutePath(), the same approach the script plugins already use (e.g. Python PYTHONPATH). Docker behavior is unchanged; the Kubernetes runner is fixed.

edit: I didnt add test because I do not know how I should test within a K8S runner context, if you have any example I'll try to write one


jymaire (approved):
Nit: if a unit test to reproduce the bug is possible, it would be nice

PR #672, opened 2026-08-13
feat(bigquery): add RunTransferConfig for BigQuery Data Transfer Service

### LINKED ISSUE #671: BigQuery Data Transfer Service: run a transfer or scheduled query and wait
### Feature description

plugin-gcp has no Data Transfer Service support: no DTS task, model, or client usage anywhere in the repo. Scheduled transfers and BigQuery scheduled queries are therefore run through generic Python tasks, roughly 25 lines of container image, pip install, service account wiring, timeout and retry per transfer. One user runs 8 of them. Orchestra covers the same job with a single required parameter.

In the DTS API a scheduled query is a transfer config, so one task covers both.

## Goal
Trigger a DTS transfer config by resource name, wait for it, emit an output asset.

## Suggested scope
- Takes a `transferConfig` resource name (`projects/{project}/locations/{location}/transferConfigs/{config}`) and nothing else that is mandatory.
- Waits for a terminal state by default.
- Re-attach before triggering: list PENDING and RUNNING runs first, adopt one if found, only then start a new run. A retry resumes, never duplicates. Note that the dbt Cloud reattach mechanism cannot be reused here: it tags dbt's free-text `cause` field, which DTS has no equivalent of.
- Poll deadline is wall-clock, not iteration count. A run parked in PENDING must still time out with a real message.
- Poll interval configurable; task timeout can exceed the task runner's pod timeout so the orderly failure path fires before the pod is killed.
- Emits a `Dataset` or `Table` output asset for the destination.

## Notes on cost
`build.gradle` already pulls `google-cloud-bigquery` under the platform-managed BOM, so adding `google-cloud-bigquerydatatransfer` is a one-line dependency change with the version already pinned. Caveat: the GCP libraries-bom wants protobuf 4.33.2 while kestra core pins 3.25.x (see the comment at `build.gradle:44`), so the DTS client inherits that constraint.

## Out of scope
Creating or editing transfer configs.

## Context
Two production incidents at n8n trace to the missing behaviours above: a worker loss on 2026-07-30 re-fired a non-idempotent trigger, and a 1h pod timeout killed a healthy 1h33m Google-side wait on 2026-08-03. Orchestra reference: https://docs.getorchestra.io/docs/integrations/gcp_big_query/gcp_bq_run_scheduled_query_job

Malaydewangan09:
## Summary

- Adds `io.kestra.plugin.gcp.bigquery.RunTransferConfig`, a task that starts a manual run of an existing BigQuery Data Transfer Service transfer config (scheduled query or otherwise) by its `transferConfigName` resource name, waits for it to reach a terminal state, and emits the destination as a data-lineage asset (Enterprise Edition)
- Adds a shared `AbstractDataTransfer` base class holding the DTS client factory, following the plugin's connection-abstraction convention
- Adds `google-cloud-bigquerydatatransfer` as a one-line, BOM-managed dependency in `build.gradle`

## Requirement-by-requirement mapping to the issue

- **Trigger a transfer config by resource name, nothing else mandatory**: `transferConfigName` is the only required property.
- **Wait for a terminal state by default**: `wait` defaults to `true`, polling until `SUCCEEDED`, `FAILED`, or `CANCELLED`.
- **Re-attach before triggering**: `findInFlightRun` lists `PENDING`/`RUNNING` runs for the config first and adopts the most recent one instead of starting a duplicate. This does not reuse the dbt Cloud `cause`-tagging mechanism (DTS has no equivalent free-text field), it lists in-flight runs directly.
- **Wall-clock poll deadline, not iteration count**: implemented with `Await.until(callable, pollInterval, maxDuration)`; a run parked in `PENDING` past `maxDuration` times out with a message naming the run and the last observed state.
- **Configurable poll interval, timeout can exceed the task runner's pod timeout**: `pollInterval` and `maxDuration` are both independent `Property<Duration>` inputs.
- **Emit a `Dataset`/`Table` output asset for the destination**: `emitDestinationAsset` emits a `Custom` asset of type `io.kestra.plugin.ee.assets.Table`, built from the config's destination dataset and (when the config targets a literal, non-templated destination table) the table name.
- **Out of scope (create/edit transfer configs)**: not implemented, matches the issue.

## Beyond the issue

Added a `reattach` `Property<Boolean>` input (default `true`, matching the issue's always-on re-attach behavior) to let a user opt out and always start a new run. This is an addition beyond the issue's suggested scope and can be dropped if the always-on behavior is preferred instead.

## Live validation

Validated against a real GCP project and a real BigQuery Data Transfer Service transfer config, plus a real Enterprise Edition asset emission pipeline:
- Happy path: run starts, task waits, run reaches `SUCCEEDED`, output populated
- Re-attach: a second execution against a config with a run already `PENDING`/`RUNNING` adopts it instead of starting a duplicate
- Asset lineage: destination dataset/table emitted and visible in the EE asset graph

## Test plan

- [x] `rtk test ./gradlew test` passes (203 tests, 0 failures; live-gated DTS tests self-skip without `GOOGLE_APPLICATION_CREDENTIALS` + `kestra.tasks.bigquerydatatransfer.config`)
- [x] Unconditional unit test for the `destinationTable` parsing helper (literal name, templated name, missing key, blank) runs on every CI run, no gating
- [x] Live-gated tests cover happy-path run-and-wait with asset-emission assertion (via a `TestAssetManagerFactory` test double), timeout on a tiny `maxDuration`, best-effort re-attach, and `reattach = false` starting a new run
- [x] `./gradlew shadowJar` builds

closes: https://github.com/kestra-io/plugin-gcp/issues/671


Malaydewangan09:
```yaml
id: dts_smoke
namespace: company.team

tasks:
      - id: run_a
        type: io.kestra.plugin.gcp.bigquery.RunTransferConfig
        assets:
          enableAuto: true
        projectId: kestra-unit-test
        transferConfigName: projects/214027225066/locations/europe/transferConfigs/6aacfffc-0000-2f9b-81a3-3c286d3750c2
        pollInterval: PT5S
```

Screenshot 2026-08-14 at 1 32 14 AM

Screenshot 2026-08-14 at 12 25 38 PM

PR #687, opened 2026-09-04
fix(bigquery): never re-attach to a future-scheduled transfer run

### LINKED ISSUE #686: RunTransferConfig: reattach adopts a future-scheduled PENDING run, so refresh-window transfers never start their own run
### Describe the issue

`RunTransferConfig` with the default `reattach: true` adopts a **future-scheduled** `PENDING` transfer run and reports it as "in-flight". The task then never starts the run it was asked to start, and it blocks until Google's own schedule reaches that queued run.

`findInFlightRun` lists runs in `PENDING` + `RUNNING` and keeps the candidate with the **greatest** `scheduleTime`:

https://github.com/kestra-io/plugin-gcp/blob/main/src/main/java/io/kestra/plugin/gcp/bigquery/RunTransferConfig.java#L241-L268

There is no upper bound on `scheduleTime`. `reattachMaxAge` only filters candidates that are *older* than a cutoff, so a run scheduled hours in the future is always eligible, and being the greatest `scheduleTime` it is always the one picked.

This breaks every Google first-party transfer that has a refresh window (Google Ads, GA4, YouTube). One daily schedule enqueues a chain of `PENDING` runs, one per backfill date, spread into the future. The config therefore always has queued `PENDING` runs, so the task always re-attaches and never starts a run of its own.

### Evidence

Google Ads transfer config, `schedule: every day 01:00`, 7-day refresh window. Google enqueues runs at 35-minute intervals: 01:00, 01:35, 02:10, 02:45, 03:20, 03:55, 04:30.

Task started at 03:08:16:

```
2026-09-03T03:08:16.381Z INFO Re-attaching to in-flight transfer run
  '.../transferConfigs/6a6d54ca-.../runs/6aab7bc5-...' instead of starting a new one
```

The adopted run, from the DTS API:

```
state       = SUCCEEDED
scheduleTime = 2026-09-03T04:30:00Z   <- 1h22m in the future at adoption time
startTime    = 2026-09-03T04:30:00.572Z
endTime      = 2026-09-03T04:33:45.507Z
runTime      = 2026-08-27T00:00:00Z   <- tail of the refresh window, not today's data
```

Same thing the day before, so it is not a one-off. Task started 03:06:51, adopted the run scheduled for 04:30, and only returned at 04:33:58:

```
2026-09-02T03:06:51.221Z INFO Re-attaching to in-flight transfer run '.../runs/6a9f6484-...'
2026-09-02T04:33:58.826Z DEBUG Emitted destination asset '<project>.google_ads'
```

adopted run `6a9f6484`: `scheduleTime 2026-09-02T04:30:00Z`, `runTime 2026-08-26T00:00:00Z`.

So the task idles for about 85 minutes every day, then reports success for a run that loaded a 7-day-old date slice. With a shorter `maxDuration` it fails on timeout instead.

### Expected behavior

A run scheduled in the future is queued, not in-flight. `findInFlightRun` should ignore any candidate whose `scheduleTime` is after `now`.

That keeps the point of `reattach` intact. A run started by the task through `startManualTransferRuns` gets `scheduleTime` = the requested run time, which is `Instant.now()` at the time of the call:

```
run 6aac5a25-...  scheduleTime 2026-09-03T03:01:37.862Z  (started by the task at 03:01:37.7)
```

By the time a retry looks for it, that `scheduleTime` is in the past, so it is still adopted.

Suggested change in `findInFlightRun`:

- skip candidates whose `scheduleTime` is after `now`;
- prefer a `RUNNING` candidate over a `PENDING` one;
- among the remaining candidates, keep the greatest `scheduleTime` as today.

A doc note that `reattach` interacts with refresh-window transfers would help too.

### Environment

- Kestra 2.0 EE, Kestra Cloud
- plugin-gcp 2.11.0
- Transfer type: `dataSourceId: google_ads`, `schedule: every day 01:00`, refresh window enabled
- Task config: `reattach` left at its default, `pollInterval: PT30S`, `maxDuration: PT2H`, `timeout: PT3H`

Malaydewangan09:
Closes #686

`findInFlightRun` kept the candidate with the greatest `scheduleTime` and had no upper bound on it, so a run scheduled in the future always won. `reattachMaxAge` only trims candidates that are too old.

Transfers with a refresh window (Google Ads, GA4, YouTube) permanently keep a chain of `PENDING` runs queued ahead of now, one per backfill date. The config therefore always had a candidate, so the task adopted one on every execution and never started its own. Reported effect: the task idled ~85 minutes, then reported success for a run that loaded a 7-day-old date slice.

## Changes

- Skip a candidate whose `scheduleTime` is beyond now **only while it is still `PENDING`**. A run DTS reports as `RUNNING` is in-flight whatever its schedule time, and gating on state avoids dropping every in-flight run when a worker's clock lags Google by more than the grace. Thanks @jymaire.
- Prefer a `RUNNING` candidate over a `PENDING` one, falling back to latest `scheduleTime` within a state.
- Reject a negative `reattachMaxAge`. It pushed the staleness cutoff ahead of the future cutoff, so every candidate failed the freshness test and reattach silently never fired.
- Parse `scheduleTime` once per candidate, and derive both cutoffs from one `Instant.now()` read.
- One `INFO` line when queued runs are skipped. A count, not one line per run, since a refresh-window config can queue many.

## Why the cutoff is `now + 30s` and not `now`

The issue suggests a strict `scheduleTime > now` skip. Its own evidence argues for a tolerance: `run 6aac5a25` was requested at `03:01:37.7` and came back with `scheduleTime 03:01:37.862Z`, so DTS stamps a run slightly ahead of the call that created it. Under a strict comparison a fast retry would refuse to adopt the run the task itself just started and would start a duplicate, which is the failure `reattach` exists to prevent.

30s is ~200x the observed drift and ~70x below the 35-minute spacing of refresh-window runs. Kept as a private constant rather than a property: it is a clock-drift tolerance, not something a user should have to reason about.

## Tests

Six added to `RunTransferConfigWireMockTest`, 14 pass in the suite.

- `reattachIgnoresFutureScheduledRunAndStartsNewOne` covers the reported scenario.
- `reattachAdoptsRunningRunScheduledInTheFuture` locks in that the cutoff never reaches a `RUNNING` run.
- `reattachPrefersRunningOverPending` gives the `PENDING` candidate the later `scheduleTime`, so schedule-time-only ordering picks the wrong run.
- `reattachAdoptsRunScheduledWithinGrace` guards the drift case against over-correction.
- `reattachMaxAgeAndFutureCutoffBothApply` covers the two bounds interacting.
- `negativeReattachMaxAgeThrows` covers the guard.

Verified the behavioral ones are not vacuous: each fails with its corresponding change reverted.

The new tests build `scheduleTime` from `Instant.now()` rather than using literals. The existing stubs hardcode dates like `2026-01-01T00:00:00Z`, correct today but silently inverting once wall-clock time crosses them. Left alone here.

## Follow-up, not in this PR

`TransferRun` has a `schedule` field, documented as empty for manually-started runs and populated for runs the config's schedule created. Filtering on it would identify the task's own runs deterministically and remove the time heuristic. Not done here because it is a policy change rather than a drop-in: it would refuse to adopt the config's own scheduled run while that run is `RUNNING`, firing a second concurrent load, and a console backfill via `requestedTimeRange` creates many manual runs at once, which reintroduces the same shape through another door.

## Note for release notes

This changes the behavior of an existing default. Users on `reattach: true` with a refresh-window transfer will see `reattached` go from `true` to `false` and the task start its own run.

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/RunTransferConfig.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/RunTransferConfig.java
@@ -165,6 +165,10 @@ public class RunTransferConfig extends AbstractDataTransfer implements RunnableT
         var rWait = runContext.render(this.wait).as(Boolean.class).orElse(true);
         var rReattach = runContext.render(this.reattach).as(Boolean.class).orElse(true);
         var rReattachMaxAge = runContext.render(this.reattachMaxAge).as(Duration.class).orElse(null);
+        // A negative value pushes the staleness cutoff into the future, which would silently reject every candidate.
+        if (rReattachMaxAge != null && rReattachMaxAge.isNegative()) {
+            throw new IllegalArgumentException("`reattachMaxAge` must not be negative, got " + rReattachMaxAge);
+        }
         var rPollInterval = runContext.render(this.pollInterval).as(Duration.class).orElse(Duration.ofSeconds(15));
         var rMaxDuration = runContext.render(this.maxDuration).as(Duration.class).orElse(Duration.ofHours(1));
 
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/RunTransferConfig.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/RunTransferConfig.java
@@ -100,6 +100,9 @@ public class RunTransferConfig extends AbstractDataTransfer implements RunnableT
         description = """
             When `true` (default), the task first looks for a run of this config that is already \
             pending or running and adopts it, so a retried execution never starts a duplicate run. \
+            A run whose schedule time is in the future is never adopted: it is queued, not in-flight. \
+            This matters for transfers with a refresh window (Google Ads, GA4, YouTube), where Google \
+            permanently keeps a chain of pending runs queued ahead of now, one per backfill date. \
             By default any in-flight run is adopted regardless of its age; set `reattachMaxAge` to opt into \
             treating an old in-flight run as stale so a fresh run is started instead. \
             Set to `false` to always start a new run."""
@@ -238,14 +241,28 @@ public class RunTransferConfig extends AbstractDataTransfer implements RunnableT
         return output(finalRun.getName(), finalRun.getState().name(), reattached, config.getDestinationDatasetId(), rTransferConfigName);
     }
 
+    // DTS stamps scheduleTime a fraction of a second after the startManualTransferRuns call that requested
+    // the run, so a bare "now" cutoff would make a fast retry refuse to adopt the run this task just started.
+    private static final Duration REATTACH_FUTURE_GRACE = Duration.ofSeconds(30);
+
     private static TransferRun findInFlightRun(DataTransferServiceClient client, String configName, Duration reattachMaxAge, Logger logger) {
+        var now = Instant.now();
+
         // reattachMaxAge is opt-in and deliberately independent of maxDuration: maxDuration is also this
         // task's own wait budget, so using it as the staleness cutoff would make a run that is still
         // legitimately in-flight after a first attempt timed out look "stale" on the retry, causing a
         // duplicate run to be started -- defeating the whole purpose of reattach.
-        var cutoff = reattachMaxAge == null ? null : Instant.now().minus(reattachMaxAge);
+        var staleCutoff = reattachMaxAge == null ? null : now.minus(reattachMaxAge);
 
-        TransferRun mostRecent = null;
+        // A run scheduled in the future is queued, not in-flight. Transfers with a refresh window (Google
+        // Ads, GA4, YouTube) permanently keep a chain of such PENDING runs, one per backfill date, so
+        // without this bound the config always has a candidate and the task never starts a run of its own.
+        var futureCutoff = now.plus(REATTACH_FUTURE_GRACE);
+
+        TransferRun best = null;
+        Instant bestScheduleTime = null;
+        var queuedSkipped = 0;
+
         for (TransferRun candidate : client.listTransferRuns(
             ListTransferRunsRequest.newBuilder()
                 .setParent(configName)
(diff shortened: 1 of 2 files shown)


src/main/java/io/kestra/plugin/gcp/bigquery/RunTransferConfig.java
```diff
@@ -238,36 +246,59 @@ Output run(RunContext runContext, DataTransferServiceClient client) throws Excep
...
-                        candidate.getName(), scheduleTime, reattachMaxAge
-                    );
-                    continue;
-                }
-            }
-            if (mostRecent == null || Timestamps.compare(candidate.getScheduleTime(), mostRecent.getScheduleTime()) > 0) {
-                mostRecent = candidate;
+            var scheduleTime = Instant.ofEpochMilli(Timestamps.toMillis(candidate.getScheduleTime()));
+
+            if (scheduleTime.isAfter(futureCutoff)) {
```
jymaire: The future-schedule skip is applied to `RUNNING` candidates too, but a run DTS reports as `RUNNING` is in-flight by definition, whatever its `scheduleTime`. Two concrete failure modes:

1. **Worker clock skew** — `now` is the Kestra worker's clock while `scheduleTime` comes from Google. If the worker lags Google by more than the 30s `REATTACH_FUTURE_GRACE` (unsynced container, VM resume), every genuinely in-flight run looks "queued", reattach silently never fires, and a retried execution starts a duplicate run against the same config — exactly what reattach exists to prevent.
2. **Backfill / refresh-window run already started** — DTS has moved it to `RUNNING` while its `scheduleTime` still points at a future slice; it gets dropped and the task starts a second concurrent run.

Restricting the skip to `state == PENDING` keeps the reported bug fixed (the queued chain is all PENDING) while removing both.

Malaydewangan09: Good catch, applied in 653adef. The skip is now gated on `state == PENDING`.

You're right that this contradicted the fix's own rationale: I justified it as "future-scheduled means queued, not in-flight", then applied it to a state that is in-flight by definition. Clock skew is the failure mode that actually bites, since a worker lagging Google by more than the grace would have dropped every in-flight run and turned reattach off exactly when it matters.

Added `reattachAdoptsRunningRunScheduledInTheFuture` to lock it in (RUNNING, `scheduleTime` +1h, must still be adopted). It fails against the previous commit. The reported bug is untouched, the queued chain is all PENDING.

On your second point, I couldn't find evidence DTS moves a run to RUNNING before its `scheduleTime` (the issue's own trace has `startTime` 572ms *after* `scheduleTime`), so I think that one is theoretical. The change covers it either way.

jymaire (approved):

PR #688, opened 2026-09-04
fix(bigquery): derive the job id from the taskrun so a resubmit cannot run it twice

### LINKED ISSUE #674: Re-attach: set a deterministic BigQuery job id so a resubmit cannot run the job twice
### Problem

BigQuery tasks submit a job and wait for it, but the job id is generated fresh on every submission, so a worker-loss resubmit starts a **second** job doing the same work. For a `Query` that writes, or a `LoadFromGcs`, that is duplicated data as well as duplicated cost.

`BigQueryService.jobId()` builds the id with only project and location and never calls `setJob(...)`, so BigQuery assigns a random id each time:

```java
return JobId.newBuilder()
    .setProject(...)
    .setLocation(...)
    .build();
```

Affected callers: `bigquery/Query.java`, `bigquery/Copy.java`, `bigquery/LoadFromGcs.java`.

### Proposed

BigQuery supports a **caller-supplied job id**, and that is its documented idempotency mechanism: submitting the same job id twice returns the existing job instead of creating a second one.

Derive the job id deterministically from the taskrun id, which is preserved across a worker-loss resubmit (`TaskRun.onRunningResend()` copies the taskrun and only touches the attempts list). A resubmit then collides with the original job and adopts it rather than duplicating it.

This is the same shape as the Databricks `idempotency_token` fix: server-side dedup, no KV state, no polling for an existing run.

### Notes

- Job ids must be unique per project **per 24h**, so the derived id needs enough entropy that a legitimate re-run after retention does not collide. Including the execution id alongside the taskrun id is the obvious approach; retries that deliberately create a new execution then get a new job, which is correct.
- Adopting a job started by a different execution or outside Kestra is out of scope here; deterministic ids solve the duplicate-submission case only.

### Related

- https://github.com/kestra-io/kestra/pull/17995 — the shared contract, for the restart-survival half
- `bigquery/RunTransferConfig` in this same plugin already implements external-run adoption, so the two mechanisms should be described consistently in the docs

### Environment

- Kestra Version: develop

Malaydewangan09:
Closes #674

`BigQueryService.jobId()` set only project and location and never called `setJob(...)`, so BigQuery assigned a random id on every submission. A worker-loss resubmit therefore started a **second** job doing the same work, with the original still running and invisible to Kestra. For a `Query` that writes, or a load, that is duplicated data as well as duplicated cost.

Deriving the id from the execution and taskrun makes the resubmit collide with the job the lost worker started, and BigQuery's caller-supplied job id is its documented dedup mechanism.

## Changes

- `BigQueryService.jobId()` now sets `kestra_<executionId>_<taskRunId>`, read from `runContext.taskRunInfo()`.
- New `BigQueryService.createOrAdoptJob()` handles a taken id: fetch the existing job and wait on it.
- Applied to `Query`, `Copy`, `LoadFromGcs`, and `ExtractToGcs`.
- A line in each task's description, so the behavior is discoverable next to `RunTransferConfig`'s `reattach`.

## Two departures from the issue

**BigQuery does not return the existing job.** The issue says a duplicate id "returns the existing job instead of creating a second one". It returns `409 ALREADY_EXISTS`. The dedup is real, but the caller gets an exception, not the job. Implemented literally, a resubmit would fail with "Already Exists" rather than adopting, trading duplicated data for a hard failure. So the conflict is caught and the job fetched explicitly.

**No attempt number in the id.** The issue proposes attempt entropy so a retry gets a fresh job. `TaskRunInfo` does not expose `attemptsCount`, and it is not clearly stable across a resubmit: `onRunningResend()` appends an attempt when the list is empty but replaces in place when it is not.

`plugin-dbt`'s `TriggerRun` already solved this without attempt entropy, keying on the taskrun alone and deciding from the found run's state:

> Adopt an in-flight or succeeded run, but never an Error or Cancelled one, so a genuine failure lets the retry or restart re-trigger instead of re-reporting the old failure.

One part does not carry over. dbt Cloud assigns its own run ids, so it can simply trigger a new run. BigQuery reserves the id we supply, so a failed job's id is unusable forever. Hence the extra branch: on a conflict with a job that is `DONE` and carries an error, resubmit with a random id. Without it, retries on these tasks would be permanently broken.

**Execution id dropped from the derived id.** The issue asked for execution id plus taskrun id. At the Kestra version this builds against, `taskRunInfo().executionId()` returns null even for a real task run, so concatenating it produced a literal `null` segment and a null check on it would have disabled the feature entirely. The id keys on `taskRunId` alone, which is an `IdUtils.create()` value already carrying full entropy, so the issue's uniqueness requirement is met without it. The execution id only added traceability, which is not worth depending on a field that is not populated.

## Beyond the issue's list

`ExtractToGcs` submitted `JobInfo.of(configuration)` with no job id at all and has the same defect. Included rather than left as the only one still broken.

`Load` is **not** covered. It creates its job through a `TableDataWriteChannel` rather than `connection.create`, so supplying an id there is a different change with different failure modes. Worth a separate issue.

## Tests

`BigQueryJobIdTest`, 7 tests, Mockito against a mock `BigQuery`, no credentials needed. 48 pass across the credential-free BigQuery suites.

- id is derived from the taskrun, stable across calls, distinct across taskruns
- a taken id whose job is running is adopted, and `create` is called exactly once
- a taken id whose job failed falls back to a random id, asserted on the resubmitted `JobInfo`
- a conflict whose job cannot be fetched is rethrown
- a non-409 error is not swallowed and never triggers a fetch

Verified the id tests are not vacuous: all three fail with `setJob(...)` removed. The first version of the stability test passed against the reverted code, since two absent job ids compare equal, so it now compares the job segment and asserts non-null.

## Note for review

This is always-on rather than an opt-in property. `plugin-dbt` made its `reattach` opt-in because it mutates a user-authored `cause` string, so upgrading changes nothing for existing flows. A BigQuery job id is not user-authored, so the same argument does not apply here, but it is a judgment call worth confirming.

### DIFF
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
@@ -1,5 +1,6 @@
 package io.kestra.plugin.gcp.bigquery;
 
+import java.net.HttpURLConnection;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -8,10 +9,13 @@ import java.util.Objects;
 
 import org.slf4j.Logger;
 
+import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryError;
 import com.google.cloud.bigquery.Job;
 import com.google.cloud.bigquery.JobException;
 import com.google.cloud.bigquery.JobId;
+import com.google.cloud.bigquery.JobInfo;
+import com.google.cloud.bigquery.JobStatus;
 import com.google.cloud.bigquery.TableId;
 
 import io.kestra.core.exceptions.IllegalVariableEvaluationException;
@@ -20,11 +24,59 @@ import io.kestra.core.runners.RunContext;
 public class BigQueryService {
     private static final String UNKNOWN_REASON = "unknown";
 
+    // BigQuery reserves a caller-supplied job id, so deriving it from the taskrun makes a worker-loss
+    // resubmit collide with the job the lost worker started instead of running the same work twice.
     public static JobId jobId(RunContext runContext, AbstractBigquery abstractBigquery) throws IllegalVariableEvaluationException {
+        var taskRun = runContext.taskRunInfo();
+
         return JobId.newBuilder()
             .setProject(runContext.render(abstractBigquery.getProjectId()).as(String.class).orElse(null))
             .setLocation(runContext.render(abstractBigquery.getLocation()).as(String.class).orElse(null))
+            .setJob("kestra_" + taskRun.executionId() + "_" + taskRun.taskRunId())
             .build();
+    }
+
+    // Same project and location with no job id, so BigQuery assigns a random one as it did before.
+    private static JobId randomJobId(JobId jobId) {
+        return JobId.newBuilder()
+            .setProject(jobId.getProject())
+            .setLocation(jobId.getLocation())
+            .build();
+    }
+
+    /**
+     * Submits the job, adopting the existing one when its deterministic id is already taken. A job that
+     * already failed has burnt its id for good, so that case falls back to a fresh random id and lets the
+     * retry make progress rather than re-reporting the old failure.
+     */
+    public static Job createOrAdoptJob(BigQuery connection, JobInfo jobInfo, Logger logger) {
+        try {
+            return connection.create(jobInfo);
+        } catch (com.google.cloud.bigquery.BigQueryException e) {
+            if (e.getCode() != HttpURLConnection.HTTP_CONFLICT) {
+                throw e;
+            }
+
+            var existing = connection.getJob(jobInfo.getJobId());
+            if (existing == null) {
+                throw e;
+            }
+
+            var status = existing.getStatus();
+            if (status != null && status.getState() == JobStatus.State.DONE && status.getError() != null) {
+                logger.info("Job '{}' already ran and failed, starting a new one", jobInfo.getJobId().getJob());
+
+                return connection.create(
+                    JobInfo.newBuilder(jobInfo.getConfiguration())
+                        .setJobId(randomJobId(jobInfo.getJobId()))
+                        .build()
+                );
+            }
+
+            logger.info("Adopting job '{}' already started by this taskrun instead of submitting a duplicate", jobInfo.getJobId().getJob());
+
+            return existing;
+        }
     }
 
     public static TableId tableId(String table) {
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
@@ -27,13 +27,18 @@ public class BigQueryService {
     // BigQuery reserves a caller-supplied job id, so deriving it from the taskrun makes a worker-loss
     // resubmit collide with the job the lost worker started instead of running the same work twice.
     public static JobId jobId(RunContext runContext, AbstractBigquery abstractBigquery) throws IllegalVariableEvaluationException {
-        var taskRun = runContext.taskRunInfo();
-
-        return JobId.newBuilder()
+        var builder = JobId.newBuilder()
             .setProject(runContext.render(abstractBigquery.getProjectId()).as(String.class).orElse(null))
-            .setLocation(runContext.render(abstractBigquery.getLocation()).as(String.class).orElse(null))
-            .setJob("kestra_" + taskRun.executionId() + "_" + taskRun.taskRunId())
-            .build();
+            .setLocation(runContext.render(abstractBigquery.getLocation()).as(String.class).orElse(null));
+
+        // A trigger evaluates outside any execution, so there is no taskrun to key on and no resubmit to
+        // deduplicate. Deriving an id there would give every poll of every trigger the same one.
+        var taskRunId = runContext.taskRunInfo().taskRunId();
+        if (taskRunId != null) {
+            builder.setJob("kestra_" + taskRunId);
+        }
+
+        return builder.build();
     }
 
     // Same project and location with no job id, so BigQuery assigns a random one as it did before.
@@ -53,7 +58,8 @@ public class BigQueryService {
         try {
             return connection.create(jobInfo);
         } catch (com.google.cloud.bigquery.BigQueryException e) {
-            if (e.getCode() != HttpURLConnection.HTTP_CONFLICT) {
+            // Without a job id of our own the conflict cannot be about a job this taskrun submitted.
+            if (e.getCode() != HttpURLConnection.HTTP_CONFLICT || jobInfo.getJobId().getJob() == null) {
                 throw e;
             }
 
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
@@ -63,6 +63,8 @@ public class BigQueryService {
                 throw e;
             }
 
+            // BigQuery says the id is taken but will not hand the job over, so there is nothing to adopt
+            // and nothing safe to resubmit under. Failing here duplicates no work, which is the point.
             var existing = connection.getJob(jobInfo.getJobId());
             if (existing == null) {
                 throw e;
@@ -70,7 +72,7 @@ public class BigQueryService {
 
             var status = existing.getStatus();
             if (status != null && status.getState() == JobStatus.State.DONE && status.getError() != null) {
-                logger.info("Job '{}' already ran and failed, starting a new one", jobInfo.getJobId().getJob());
+                logger.warn("Job '{}' already ran and failed, starting a new one", jobInfo.getJobId().getJob());
 
                 return connection.create(
                     JobInfo.newBuilder(jobInfo.getConfiguration())
@@ -79,7 +81,7 @@ public class BigQueryService {
                 );
             }
 
-            logger.info("Adopting job '{}' already started by this taskrun instead of submitting a duplicate", jobInfo.getJobId().getJob());
+            logger.warn("Adopting job '{}' already started by this taskrun instead of submitting a duplicate", jobInfo.getJobId().getJob());
 
             return existing;
         }
@@ -100,9 +102,7 @@ public class BigQueryService {
         if (job == null) {
             throw new IllegalArgumentException("Job no longer exists");
         } else if (job.getStatus() == null) {
-            // A submission answered with a job reference and no status yet has nothing to report. The
-            // waitFor that follows establishes the real state, so a failure cannot slip through here.
-            return;
+            throw new IllegalStateException("Job '" + job.getJobId().getJob() + "' has no status to report");
         } else if (job.getStatus().getError() != null) {
             ArrayList<BigQueryError> errors = new ArrayList<>();
             if (job.getStatus().getError() != null) {
--- a/src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
+++ b/src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
@@ -99,6 +99,10 @@ public class BigQueryService {
     public static void handleErrors(Job job, Logger logger) throws BigQueryException {
         if (job == null) {
             throw new IllegalArgumentException("Job no longer exists");
+        } else if (job.getStatus() == null) {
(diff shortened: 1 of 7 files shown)


src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
```diff
@@ -20,13 +24,67 @@
...
+        try {
+            return connection.create(jobInfo);
+        } catch (com.google.cloud.bigquery.BigQueryException e) {
+            // Without a job id of our own the conflict cannot be about a job this taskrun submitted.
+            if (e.getCode() != HttpURLConnection.HTTP_CONFLICT || jobInfo.getJobId().getJob() == null) {
+                throw e;
+            }
+
+            var existing = connection.getJob(jobInfo.getJobId());
+            if (existing == null) {
```
fdelbrayelle, line 69: 🟡 [Correctness] Eventual-consistency race on adopt
Problem: if create() throws 409 but getJob(jobInfo.getJobId()) has not yet caught up and returns null, the original conflict e is rethrown. In AbstractBigquery.waitForJob, that lands with job == null at the point of failure, so retryable = jobId != null && ... evaluates to false — the outer Failsafe retry never gets a chance, and a transient visibility lag becomes a hard task failure instead of the intended adopt.
Fix: consider a short bounded retry/backoff around the getJob() lookup before giving up, or explicitly document this as an accepted fail-safe (non-duplicating) trade-off.

src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
```diff
@@ -20,13 +24,67 @@
...
+                throw e;
+            }
+
+            var status = existing.getStatus();
+            if (status != null && status.getState() == JobStatus.State.DONE && status.getError() != null) {
+                logger.info("Job '{}' already ran and failed, starting a new one", jobInfo.getJobId().getJob());
```
fdelbrayelle: 🟡 [Guidelines] Log level for recovery-path events
Problem: both this line and line 82 use logger.info for adopting an orphaned job / discarding a burnt job id — both signal something abnormal happened upstream (worker loss, or a taskrun retrying a burnt id), not routine progress.
Fix: consider logger.warn so operators notice a recovery occurred instead of it blending into normal INFO-level job-lifecycle logging.

src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
```diff
@@ -41,6 +99,10 @@ public static TableId tableId(String table) {
     public static void handleErrors(Job job, Logger logger) throws BigQueryException {
         if (job == null) {
             throw new IllegalArgumentException("Job no longer exists");
+        } else if (job.getStatus() == null) {
+            // A submission answered with a job reference and no status yet has nothing to report. The
+            // waitFor that follows establishes the real state, so a failure cannot slip through here.
```
fdelbrayelle: 🟡 [Correctness] Null-status tolerance and dry run
Problem: this early return treats a bare, null-status job reference as 'nothing to report yet', relying on the caller's subsequent job.waitFor() to re-establish real state. AbstractBigquery.waitForJob skips waitFor() entirely when dryRun is true, and dry-run submissions now also carry a caller-supplied job id (the same trigger the third commit found causes bare job references). If BigQuery ever answers a dry-run insert with a null-status job instead of throwing synchronously, a real dry-run error would go unreported.
Fix: add a regression test that drives a dry-run Query with intentionally invalid SQL through the deterministic-job-id path, to confirm the error still surfaces via the exception path.

src/test/java/io/kestra/plugin/gcp/bigquery/BigQueryJobIdTest.java
```diff
@@ -0,0 +1,211 @@
...
+    void shouldGiveDifferentJobIdsToDifferentTaskruns() throws Exception {
+        var task = task();
+        var first = TestsUtils.mockRunContext(runContextFactory, task, ImmutableMap.of());
+        var second = TestsUtils.mockRunContext(runContextFactory, task, ImmutableMap.of());
+
+        assertThat(BigQueryService.jobId(first, task).getJob(), not(BigQueryService.jobId(second, task).getJob()));
+    }
+
+    @Test
+    void shouldAdoptTheRunningJobWhenTheIdIsAlreadyTaken() {
```
fdelbrayelle, line 89: 🟡 [Tests] Only the RUNNING branch of adopt is covered
Problem: job(null) always yields JobStatus.State.RUNNING, so this test never exercises adopting an existing job that is already DONE and successful (status.getError() == null) — the other real outcome of the same if condition in BigQueryService, and the boundary that is easiest to get subtly wrong.
Fix: add a sibling test with a mocked job(DONE state, no error) asserting it is adopted the same way as the RUNNING case.

src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
```diff
@@ -20,13 +24,67 @@
...
     public static JobId jobId(RunContext runContext, AbstractBigquery abstractBigquery) throws IllegalVariableEvaluationException {
-        return JobId.newBuilder()
+        var builder = JobId.newBuilder()
             .setProject(runContext.render(abstractBigquery.getProjectId()).as(String.class).orElse(null))
-            .setLocation(runContext.render(abstractBigquery.getLocation()).as(String.class).orElse(null))
+            .setLocation(runContext.render(abstractBigquery.getLocation()).as(String.class).orElse(null));
+
+        // A trigger evaluates outside any execution, so there is no taskrun to key on and no resubmit to
+        // deduplicate. Deriving an id there would give every poll of every trigger the same one.
+        var taskRunId = runContext.taskRunInfo().taskRunId();
```
fdelbrayelle, line 36: 🟡 [Business Requirements] Execution id dropped from the derived id, not called out as a departure
Problem: issue #674 explicitly asked for execution id + taskrun id for entropy against 24h retention collisions. This line keys on taskRunId alone, for every caller, not only the trigger fallback — a real narrowing of the issue's proposal that isn't listed in the PR description's 'Two departures from the issue' section. Functionally very likely fine since taskRunId is a standalone IdUtils.create() value already carrying full entropy, but the departure deserves the same explicit call-out the other two get.
Fix: add a short note to the PR description explaining why taskRunId alone already satisfies the issue's uniqueness requirement.

fdelbrayelle (approved):
## Kestra Plugin Code Review

### Business Requirements — met, with one undocumented departure
Fixes #674: `BigQueryService.jobId()` now derives a caller-supplied job id from `taskRunInfo().taskRunId()`, and `createOrAdoptJob()` adopts the job on a 409 conflict, falling back to a random id when the taken id belongs to an already-failed job. Applied to `Query`, `Copy`, `LoadFromGcs`, and (beyond the issue's list) `ExtractToGcs`. `Load` is deliberately excluded with a sound rationale (goes through `TableDataWriteChannel`, not `connection.create`).

One thing not called out in the PR's own 'Two departures from the issue' section: the issue explicitly recommended keying on **execution id + taskrun id** ('the derived id needs enough entropy that a legitimate re-run after retention does not collide... Including the execution id alongside the taskrun id is the obvious approach'). The second commit drops execution id entirely, everywhere, not just for the trigger fallback — `jobId()` now keys on `taskRunInfo().taskRunId()` alone. Verified against `DefaultRunContext.taskRunInfo()`: `taskRunId` is a freestanding `IdUtils.create()` value, not scoped by execution, so it already carries full entropy on its own and the omission is very likely harmless — but it silently narrows a requirement the issue was explicit about, and that narrowing isn't surfaced to reviewers the way the other two departures are.

### Kestra Guidelines — 1 finding
🟢 **Nit** — `@Schema(description = ...)` on `Copy`, `Query`, `LoadFromGcs`, `ExtractToGcs` stays a single-line string after the new sentence was appended; the guideline calls for Java `"""` blocks for `@Schema(description = ...)`. Pre-existing file convention, not introduced by this PR, so not blocking.

### Security (OWASP Top 10:2025 + KPS) — ✅ No issues found
No secrets, no new dependencies, no HTTP client bypass, no injection surface — the job id is built from an internally generated taskrun id, not user input.

### Performance — ✅ No issues found
No unbounded allocation, no resource leaks, no thread-safety issues; `createOrAdoptJob`/`randomJobId` are stateless and cheap.

### Additional non-blocking observations (see inline comments)
- Eventual-consistency race in `createOrAdoptJob`: if `getJob()` right after a 409 returns `null`, the original conflict is rethrown and `AbstractBigquery.waitForJob`'s outer retry treats it as non-retryable, turning a narrow race into a hard failure instead of a duplicate.
- `logger.info` for both the 'adopting an orphaned job' and 'discarding a failed job's id' paths — arguably `warn`, since both indicate a worker-loss / burnt-id recovery, not routine progress.
- `handleErrors`' new null-status tolerance is only re-verified by the follow-up `job.waitFor()`, which is skipped for `dryRun`. Worth an explicit test confirming a dry-run SQL error still surfaces correctly now that dry runs also carry a caller-supplied job id.
- Test coverage: the adopt path is only tested for `RUNNING`, not for an already-`DONE`-and-successful job — the other real outcome of the same condition, and the trickiest boundary to get right.

Verified independently against `kestra/core`'s `DefaultRunContext.taskRunInfo()` that `taskRunId()` returns `null` (not an NPE) when there's no `taskrun` variable, confirming the second commit's trigger-context fix is correct, and that `bigquery/Trigger.java` does build a `Query` and call `task.run()` from a trigger-evaluation `RunContext` with no taskrun — so the scenario the second commit fixes ('every poll of every trigger got the same id') is real and was a legitimate near-miss caught before merge.

**Verdict:** APPROVE

src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
Malaydewangan09: Documented as the accepted trade-off rather than retried, in 1674444.

Agreed on the mechanism: `job == null` and `lastJobId == null` at that point, so `retryable` is false and it becomes a hard failure. I looked at a bounded retry around the lookup and did not take it, because the likeliest reason `getJob` comes back empty is not visibility lag but a location mismatch. `jobInfo.getJobId()` carries whatever location the task declared, which is often none, while the insert resolved one from the referenced datasets. Retrying the same lookup would not find it either.

Failing there duplicates no work, which is the property worth keeping. Comment added saying so.

src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
Malaydewangan09: Both moved to `warn` in 1674444. Agreed, they mark a recovery from worker loss or a burnt id, not routine progress.

src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
Malaydewangan09: Real gap, fixed in 1674444, and the blanket guard was the wrong shape.

You are right that `waitForJob` skips `waitFor()` on a dry run, so `handleErrors` is its only check and a bare job reference would have been reported as success. The tolerance now sits at the submission check in `AbstractBigquery`, where the poll that follows still establishes the real state. `handleErrors` refuses a job it cannot report on, so a dry run answered without a status now fails loudly instead of passing.

Test added on `handleErrors` directly. `shouldSkipLookbackOnDryRunRetry` already covers a dry run whose status carries an error and still passes.

src/test/java/io/kestra/plugin/gcp/bigquery/BigQueryJobIdTest.java
Malaydewangan09: Added `shouldAdoptAnAlreadyFinishedJobWhenTheIdIsAlreadyTaken` in 1674444, asserting a DONE job with no error is adopted and `create` is called once. `job()` takes an explicit state now so the two branches are separable.

src/main/java/io/kestra/plugin/gcp/bigquery/BigQueryService.java
Malaydewangan09: Added to the PR description as a third departure.

Worth recording why it is not only a trigger fallback: at the Kestra version this builds against, `taskRunInfo().executionId()` returns null even for a real task run, so the original concatenation was producing a literal `null` segment. A null check on it would have disabled the feature for every task rather than just triggers. `taskRunId` is an `IdUtils.create()` value on its own, so the entropy the issue wanted is already there and the execution id was only adding traceability.

PR #696, opened 2026-09-14
refactor: consume plugin-gcp-lib shared kernel

jymaire:
## What

Depend on the new shared-kernel library `plugin-gcp-lib` instead of the OSS-local copies of the GCP auth/connection core.

- Delete `io.kestra.plugin.gcp.{GcpInterface,CredentialService,AbstractTask}` — now provided by `io.kestra.plugin.gcp.shared.*` in the lib.
- Add `implementation 'io.kestra.plugin:plugin-gcp-lib:1.0.0-SNAPSHOT'` and update imports across every task package.

Behaviour-preserving move: no task ids/aliases change, and the property surface is unchanged (the `serviceAccount`/`impersonatedServiceAccount` `group = "connection"` normalisation comes from the shared kernel).

## Dependency note

This PR consumes `plugin-gcp-lib:1.0.0-SNAPSHOT`, currently wired via `mavenLocal()` (marked with a `TODO` in `build.gradle`). CI will pass once that SNAPSHOT is available in the snapshot repository; the `mavenLocal()`/`TODO` line and the SNAPSHOT pin should be swapped for the released version before merge.

part-of: https://github.com/kestra-io/kestra-ee/issues/7116

---

## Update — consumes the collapsed `CredentialService.connection` API

- The shared kernel now exposes a single `CredentialService.connection(...)` returning `GcpConnection(credentials, projectId)`; the old `credentials(...)` entry point was removed. The deprecated `io.kestra.plugin.gcp.CredentialService` shim and the bigtable/spanner/dataflow call sites are repointed onto it.
- **Correction to the description above:** the property groups are **not** normalised to `connection` here — they match plugin-gcp `main` property-by-property (`serviceAccount` → `execution`, `impersonatedServiceAccount` → `advanced`). The coordinated regroup is a separate follow-up (plugin-gcp-lib#2).
- **Shims are source-compatible only.** The deleted `AbstractTask`/`GcpInterface`/`CredentialService` remain as `@Deprecated(forRemoval)` forwarding shims: third-party code that *extends/implements* them still compiles, but `instanceof`/casts against plugin-gcp's own tasks do not. Ships as **2.12.0**, shims removed in the next major.
- `secret = true` on `impersonatedServiceAccount` now comes from the kernel and matches `bigtable`/`spanner`, so masking is uniform across all GCP tasks again.
- Added an offline `AbstractTaskInferenceTest` so this repo's own CI guards the project-id inference contract (the kernel's tests never run here).
- Fixes two pre-existing auth bugs surfaced in review: `HttpFunction` cast credentials to `ServiceAccountCredentials` (ClassCastException under impersonation → now casts to `IdTokenProvider`); `AbstractPubSub` read `projectId` before resolving credentials (inferred project id was missed → now resolves credentials first).
- Dataflow never infers `projectId`; that is a contract change tracked separately in #697, not bundled here.

_Depends on plugin-gcp-lib being released to Maven Central (its `release.yml` has not run yet); until then the dependency resolves via the `mavenLocal()` entry._


build.gradle
```diff
@@ -75,6 +75,9 @@ dependencies {
     // libs
     compileOnly group: 'dev.failsafe', name: 'failsafe'
 
+    // Shared kernel: GcpInterface / CredentialService / AbstractTask (local validation, see repositories block)
+    implementation 'io.kestra.plugin:plugin-gcp-lib:1.0.0-SNAPSHOT'
```
jymaire: ⛔ **Do not merge yet — blocked on `plugin-gcp-lib` release.**

This pins `plugin-gcp-lib:1.0.0-SNAPSHOT` and is resolved via the temporary `mavenLocal()` above (line 20, `TODO`). Before merging:
1. Release / publish `plugin-gcp-lib` so a resolvable version exists.
2. Bump this pin to the released version.
3. Remove the `mavenLocal()` line and its `TODO`.

CI will not pass until the SNAPSHOT is available in the snapshot repository. Leaving this thread open as a merge gate.

Malaydewangan09:
Reviewed alongside kestra-io/plugin-gcp-lib#1 and kestra-io/plugin-ee-gcp#208. The repoint itself is clean: 31 files, no test changes, and `git grep` finds no surviving reference to the three deleted classes. Three things block it.

**1. The dependency resolves nowhere.** `check / main` dies at `:compileJava` with `Could not find io.kestra.plugin:plugin-gcp-lib:1.0.0-SNAPSHOT`. plugin-gcp-lib has no releases, no tags, and its `release.yml` is `workflow_dispatch`-only and has never run. Worth knowing that `Java Tests Report` still reports **pass** with `0 ran`, so the check list reads as mostly green at a glance.

**2. The lib's fat jar gets re-shaded in here, and the stale copy wins.** plugin-gcp-lib currently publishes a 6 MB shaded jar (see my comment on #1). Because this repo sets `duplicatesStrategy = DuplicatesStrategy.INCLUDE`, that whole tree is copied in verbatim:

| jar | class entries | unique | duplicates |
|---|---|---|---|
| `plugin-gcp-2.9.2-SNAPSHOT.jar` (main) | 64508 | 64508 | 0 |
| `plugin-gcp-2.10.2-SNAPSHOT.jar` (main) | 64768 | 64768 | 0 |
| `plugin-gcp-2.11.3-SNAPSHOT.jar` (this PR) | 76384 | 72113 | **4271** |

161 MB to 180 MB. The duplicates are different bytecode, not copies. `dependencyInsight` here resolves google-auth **1.51.0**, the lib's bundled copy is **1.42.1**, and the 1.42.1 `GoogleCredentials.class` sits first in zip order. `PluginClassLoader` is child-first, so plugin-gcp compiles against 1.51.0 and loads 1.42.1. Same shape for google-http-client (2.1.0 vs 2.2.0) and grpc-api (1.76.2 vs 1.82.4). `verifyShadedGrpcLoadBalancers` still passes, so CI will not catch this. It also never ran on this branch, since the build died at `:compileJava`.

Fixing the lib to publish thin makes this go away.

**3. `implementation` should be `api`.** The generated POM declares the lib at `runtime` scope, but `AbstractBigquery`, `AbstractGcs`, `AbstractPubSub`, `AbstractComputeTask`, `AbstractRcs`, `AbstractBatch`, `AbstractDataForm`, `AbstractGenerativeAi`, `AbstractMonitoringTask` and `AbstractDataTransfer` all expose `io.kestra.plugin.gcp.shared.AbstractTask` in their supertype chain. Anything compiling against plugin-gcp gets "cannot access AbstractTask" unless it declares the lib itself. plugin-git and plugin-kubernetes both use `api` for their lib.

Also worth deciding before merge:

- The branch is 6 commits behind main, including #688, which changed `AbstractBigquery.java`, a file this PR also edits. No textual conflict, but the combination has never compiled.
- Three public classes are deleted with no shim. No flow YAML breaks (none of them is a `@Plugin` type) and nothing in kestra-io references them outside this repo, so the exposure is third-party Java extending `io.kestra.plugin.gcp.AbstractTask`. Either accept that and cut the next release as a major, or leave deprecated subclasses for one minor. Just make it a decision rather than letting it ship as 2.11.4.
- `release { failOnSnapshotDependencies }` will reject the SNAPSHOT coordinate anyway.

Smaller: the two Spanner files are formatting only, blank lines and import reordering with no behaviour change, which is noise in an otherwise mechanical 26-line import repoint. Run spotless separately. `mavenLocal()` was already on main, only the TODO comment is new, so just drop the comment.

Malaydewangan09:
Round 2, re-baselined onto `4ec446c`. Note for anyone else reading: this PR was force-pushed with a third commit that is not mentioned anywhere in the thread, so the diff you looked at yesterday is not the diff.

Good news first, measured rather than assumed. I rebuilt the lib from its current head, served it from a private repo, and built three shadow jars:

| | entries | dup class names | `GoogleCredentials.class` |
|---|---|---|---|
| origin/main | 73,973 | 0 | 1 |
| this head + the stale `~/.m2` fat jar | 78,448 | **4,271** | 2, at different versions |
| this head + a correctly built lib | 73,975 | **0** | 1, at 1.51.0 |

So the round-1 shading blocker is genuinely gone once the lib publishes thin. `verifyShadedGrpcLoadBalancers` runs and passes on all three. `api` is correct now, POM says `compile` scope. Spanner churn is gone, exactly two import lines. Branch is current. Annotations match the OSS originals property by property.

Now the problems.

**1. Still unresolvable.** `check / main` dies at `:compileJava` on `Could not find io.kestra.plugin:plugin-gcp-lib:1.0.0-SNAPSHOT`. Confirmed independently, `curl` gives 404 on both Central snapshots and repo1. And the lib cannot publish either, because dropping its shadow plugin broke its own CI (`Task 'shadowJar' not found`, it needs `generate-shadowjar: false`). Also `failOnSnapshotDependencies` would reject the coordinate even if it resolved.

**2. The TODO marker was removed but the temporary state was kept.** The entire build.gradle change in the second commit deletes the local-validation comment from the `mavenLocal()` line and nothing else. That was the signal, not the problem. The dependency comment at line 77 still says "see repositories block" and now points at a block that says nothing. `mavenLocal()` is also listed first, so it shadows every remote repo for every coordinate, not just this one.

**3. The ADC gate regresses this plugin.** The code being replaced has no gate on `serviceAccount`, so OSS infers `projectId` from a key file supplied through `GOOGLE_APPLICATION_CREDENTIALS`. Verified against google-auth 1.51.0 that ADC really does return a `ServiceAccountCredentials` for a `"type": "service_account"` file, and that `createScoped` preserves the type. The lib's javadoc says such a run "keeps failing explicitly ... as it did before the kernel extraction", which is false here, and the lib test locks the regression in with the same wrong comment.

Probed failure modes against the built jar. Hard NPE with no message:

- `dataproc/clusters/Create.java:163` and `Delete.java:80`, via `CreateClusterRequest.Builder.setProjectId(null)`
- `dataproc/batches/AbstractBatch.java:178`, via `LocationName.of(null, region)`, so every `dataproc.batches.*` task
- `vertexai/CustomJob.java:144`, same cause
- plus the four `compute/*` tasks that `orElseThrow()` on it directly

Degraded silently instead of failing, which is worse:

- `bigquery/AbstractBigquery.java:158`, `gcs/AbstractGcs.java:31`, `firestore/AbstractFirestore.java:48` all accept a null project and fall through to `ServiceOptions` default detection, so the run may target a different project than before
- `dataform/AbstractDataForm.java:64` formats a literal `projects/null/locations/...`

Not affected: pubsub, bigtable, spanner.

**4. The shims do not preserve what the commit message says.** Probed against this head's own shadow jar:

```
io.kestra.plugin.gcp.gcs.Upload       instanceof gcp.AbstractTask=false  gcp.GcpInterface=false
io.kestra.plugin.gcp.bigquery.Query   instanceof gcp.AbstractTask=false  gcp.GcpInterface=false
io.kestra.plugin.gcp.gcs.Trigger      instanceof gcp.AbstractTask=false  gcp.GcpInterface=false
```

Every real task extends `shared.AbstractTask`, not the shim. What the shims do preserve: third-party code that extends or implements the old types still compiles and links. What they do not: `instanceof` and casts against plugin-gcp's own tasks, and `io.kestra.plugin.gcp.CredentialService.credentials(ctx, someGcsTask)`, which used to compile and now does not. `AbstractTask.class.getDeclaredFields()` also returns empty where it returned four.

Either make the OSS abstract bases extend the shims so `instanceof` works, or say in the body that they are source-compat only and ship as a major. The `CredentialService` shim is complete, for what it is worth, the old class had exactly one public member.

**5. The lib's `metadata/index.yaml` shadows this plugin's.** With a correctly built lib there are two entries at that path and `DuplicatesStrategy.INCLUDE` keeps both. `getResourceAsStream` returns the lib's:

```
group: io.kestra.plugin.gcp.shared
name: "gcp-lib"
title: "GCP Lib"
```

`group: io.kestra.plugin.gcp / title: "Google Cloud"` is unreachable. Kestra core does not read that file at runtime, so whether this matters depends on the docs ingestion, which I could not check. `icons/plugin-icon.svg` is duplicated too but byte-identical, so harmless today and a landmine the next time either icon changes. The fix belongs in the lib: it registers no plugins, so drop `src/main/resources/{metadata,icons}` there.

**6. Dropping `secret = true` leaves this plugin internally inconsistent.** `bigtable/AbstractBigtable.java:48`, `bigtable/Trigger.java:80` and `spanner/AbstractSpanner.java:40` each declare their own `@PluginProperty(secret = true, group = "advanced")` on `impersonatedServiceAccount` and were not touched. After this PR the same field renders masked on Bigtable and Spanner tasks and as plain text everywhere else. On main it was uniformly masked. Decide once and apply it in one place.

Smaller:

- Scopes are now rendered twice per credential build, once in `sourceCredentials` and again for the impersonation path. The original rendered once and reused the list.
- `gradle.properties` was hand-edited to `2.12.0-SNAPSHOT`. Every prior bump in this repo came from `net.researchgate.release` as a `chore(version):` commit. Let the release plugin set it and pass the target version to the workflow.
- The lib exports `io.kestra:platform {strictly 1.3.19}` into the consumer graph. It resolves here only because this repo declares its own `enforcedPlatform`, which wins. Worth knowing that is load-bearing. The lib compiles against google-auth 1.42.1 and runs against 1.51.0.
- No test in this repo covers the auth path at all. `git grep` over `src/test` for the three class names returns nothing. All coverage now lives in the lib, which this repo's CI never runs, so the regression above is invisible here forever. One test asserting the inference contract for a single `AbstractTask` subclass would fix that.
- Pre-existing, but in files this PR touches: `dataflow/AbstractDataflow.java:39` calls the static `CredentialService.credentials(...)` rather than `this.credentials(...)`, so no Dataflow task ever infers `projectId`. `pubsub/AbstractPubSub.java:41` reads `projectId` before resolving credentials. `function/HttpFunction.java:97` casts to `ServiceAccountCredentials`, which is a `ClassCastException` whenever `impersonatedServiceAccount` is set.
- `Co-Authored-By` AI attribution on `4ec446c`. Strip it before merge.

jymaire:
Thanks for the thorough pass, @Malaydewangan09 — pushed two commits and rebased onto main.

**Addressed**
- **`implementation` → `api`** — agreed, `shared.AbstractTask` is in the public supertype chain. Done.
- **Spanner spotless noise** — reverted to import-only (3 files, 1 line each). These files aren't spotless-clean on `main`, so the `spotlessApply` pre-commit hook keeps re-adding the reflow; I bypassed it here to keep the repoint minimal and will do the spotless normalization as its own PR rather than smuggling it in.
- **`mavenLocal()` TODO comment** — dropped (kept `mavenLocal()`, which was already on main).
- **6 commits behind main (incl. #688)** — rebased; #688's `AbstractBigquery` change and the import repoint coexist cleanly.
- **Deleted public classes / no shim** — went with the shim route: re-added `io.kestra.plugin.gcp.{AbstractTask,CredentialService,GcpInterface}` as `@Deprecated(forRemoval = true, since = "2.12.0")` classes forwarding to `shared.*`, so third-party Java extending/using the old coordinates keeps compiling. Bumped to **`2.12.0-SNAPSHOT`** so this ships as a minor (not `2.11.4`); the shims come out in the next major. `compileJava` passes locally against the lib.

**Still open — not fixable in this repo**
- **Dependency resolves nowhere / `failOnSnapshotDependencies`** — stays blocked until `plugin-gcp-lib` is released; tracked in the merge-gate thread. Won't merge before then.
- **Fat-jar re-shading (google-auth 1.42.1 shadowing 1.51.0, same for google-http-client / grpc-api)** — the fix is on the lib side (publish thin, not shaded); following up on plugin-gcp-lib#1. Nothing to change here once the lib is thin.

jymaire:
Thanks — pushed `d22d237` (API) + `63c5915` (pre-existing bug fixes). Round-2 mapping:

1. **Unresolvable dependency** → external: needs the lib's first Maven Central release; pinned to `1.0.0-SNAPSHOT` via `mavenLocal()` until then.
2. **mavenLocal / TODO** → restored a clear comment on the `mavenLocal()` line (temporary, until the lib release) and fixed the stale dependency comment.
3. **ADC gate regressing OSS** → OSS keeps inferring (kernel default `true`); the lib javadoc/test comments are corrected to say so.
4. **Shims don't preserve `instanceof`** → accepted as source-compat-only; ships as **2.12.0** with the shims `@Deprecated(forRemoval)`, removed next major. Noted in the PR body.
5. **lib `index.yaml`/icon shadowing** → fixed in the lib: `src/main/resources/{metadata,icons}` dropped (kestra-io/plugin-gcp-lib#1 / `a650d64`).
6. **`secret = true` consistency** → the kernel now carries it on `impersonatedServiceAccount`, matching `bigtable`/`spanner`; masking is uniform again.

Smaller: scopes rendered once (lib); added `AbstractTaskInferenceTest` so OSS CI guards the inference contract; `api` scope confirmed. The two pre-existing bugs you flagged — `HttpFunction` CCE under impersonation, and `AbstractPubSub` reading `projectId` before credential inference — are fixed in `63c5915`. The Dataflow `projectId`-inference gap is a contract change, filed as #697 rather than bundled here. No AI co-author on the new commits.

Malaydewangan09:
Round 3 at 63c5915, built against the lib at a650d64 in an isolated maven repo.

`compileJava compileTestJava` is clean, zero errors, where the previous head had 5 against the new kernel API. The `CredentialService` shim forwards through `connection(...).credentials()` correctly. `AbstractTaskInferenceTest` passes 2 of 2, and it has teeth: deleting the write-back in the lib's `AbstractTask` turns `infersProjectIdFromServiceAccountKey` red.

The shadow jar is also clean now that the lib stopped shipping its resources:

```
total entries        75862   (was 75864, exactly the two removed)
duplicate .class         0
icons/plugin-icon.svg    x1
metadata/index.yaml      x1  -> group: io.kestra.plugin.gcp / name: "gcp"
```

That closes the round-1 shading blocker and the round-2 index.yaml shadowing in one go.

Thanks for picking up the two pre-existing bugs, both fixes read correctly. Some things left.

**1. Neither new fix has a regression test, and this repo has no impersonation coverage at all.** `grep -rn 'impersonat' src/test/java` returns nothing. `HttpFunctionTest`, `PublishThenConsumeTest` and the two pubsub triggers are all live-GCP and none sets `impersonatedServiceAccount`. So both fixes are load-bearing behaviour that CI cannot exercise. Two offline tests would cover it: assert `credentials(runContext).createScoped(...)` is `instanceof IdTokenProvider` under impersonation without calling `refreshAccessToken`, and assert `getProjectId()` is non-null after `createSubscription(runContext, "s", false)`.

**2. No test for ADC inference either.** `git grep ApplicationDefault -- src/test` returns nothing. That path is exactly what `inferProjectIdFromApplicationDefault()` defaulting to `true` exists to preserve for OSS, and it is the behaviour I flagged as regressed in round 2. It is now guarded by nothing on this side, and the kernel's own test uses a hand-rolled interface rather than yours.

**3. `HttpFunction`'s cast is still unguarded.** `(IdTokenProvider)` is correct for the reported case and strictly wider than before, since `ComputeEngineCredentials` also implements it and used to fail the old cast. But `UserCredentials`, which is what `gcloud auth application-default login` produces, implements neither, so a serviceAccount-less dev run still dies with a bare ClassCastException naming a type the user never configured. An `instanceof` check with an actionable message would finish the fix.

**4. The third bug I named is not fixed and the commit message does not mention it.** `dataflow/AbstractDataflow.java:39` discards `.projectId()`, and it cannot do otherwise as written since `dataflowClient` is static and takes a `DataflowConnectionInterface`. Every dataflow task renders `this.projectId` independently, and `dataflow/Trigger` does not extend `AbstractTask` at all, it re-declares its own four connection fields. Mitigating: `projectId` is `@NotNull` on all of them, so it fails validation loudly rather than making a null-project call. That is why I am not calling it a blocker. It is still an inconsistency, a service-account key that supplies the project id now works for gcs, bigquery and pubsub, and not for dataflow.

Smaller:

- `AbstractPubSub.createSubscription` now resolves credentials unconditionally, where before they were only touched inside the `if (autoCreateSubscription)` block. A caller with an existing subscription and unresolvable credentials previously got a `ProjectSubscriptionName` back without authenticating and now throws. `Consume` and `RealtimeTrigger` both build a Subscriber straight after, so blast radius is near zero, but it is a behaviour change for someone not hitting the bug. The fix itself is complete, the backfilled `projectId` is read at all three places the null used to leak.
- `AbstractTaskInferenceTest.keepsExplicitProjectId` survives the write-back deletion that its sibling catches. It only guards against a write-back that clobbers an explicit projectId, never a missing one. Worth keeping, not worth counting as coverage of the inference contract.
- The deprecated shims still do not preserve `instanceof`. 19 files import `io.kestra.plugin.gcp.shared.AbstractTask` and nothing imports the shim, so `someGcsTask instanceof io.kestra.plugin.gcp.AbstractTask` is false for every first-party task. Nothing in this round changed that and nothing was meant to, but it deserves a line in the deprecation release note. A compile-clean upgrade that silently flips a type check is the worst kind to debug.

Still blocked on the lib release, and I left four things on plugin-gcp-lib#1 that affect this repo, the impersonation project-id semantics being the one worth reading.

Malaydewangan09 (approved):
Approving. Compiles clean against the kernel at a650d64, the shim forwards correctly, and the shadow jar is back to zero duplicate classes with one copy each of `index.yaml` and the icon.

Do not merge until the lib is released and this pins the real version. CI is red on the SNAPSHOT as expected.

Follow-ups from my round-3 comment, none blocking: the two bug fixes in 63c5915 have no regression test and this repo has no impersonation or ADC coverage at all, the `IdTokenProvider` cast is still unguarded for `UserCredentials`, and dataflow still does not infer the project id. Worth a line in the deprecation release note that the shims do not preserve `instanceof`.

jymaire:
> Approving. Compiles clean against the kernel at a650d64, the shim forwards correctly, and the shadow jar is back to zero duplicate classes with one copy each of `index.yaml` and the icon.
> 
> Do not merge until the lib is released and this pins the real version. CI is red on the SNAPSHOT as expected.
> 
> Follow-ups from my round-3 comment, none blocking: the two bug fixes in [63c5915](https://github.com/kestra-io/plugin-gcp/commit/63c59158a348d4670bcd719adca3823aa2230ea2) have no regression test and this repo has no impersonation or ADC coverage at all, the `IdTokenProvider` cast is still unguarded for `UserCredentials`, and dataflow still does not infer the project id. Worth a line in the deprecation release note that the shims do not preserve `instanceof`.

Added in https://github.com/kestra-io/plugin-gcp-lib/issues/3

jymaire:
failed test seems to be flaky

